PR #16, opened 2021-11-09
Add React and GraphQL config options

OKitsMikael:
- [x] update settings.py
- [x] add react client files
- [x] update react directory structure
- [x] add GraphQL files
- [x] verify generated app works properly
- [x] add graphiql interface removed from #28


cookiecutter.json
```diff
@@ -16,8 +16,10 @@
     "Amazon SES",
     "Custom SMTP"
   ],
+  "use_graphql": "n",
   "client_app": [
     "Vue3",
+    "React",
     "None"
```
oudeismetis, line 23: If you choose React, do you also NEED graphql? (or vice versa?) Or does our React setup work without it?
Do we have (or are planning) support for graphql in Vue?

oudeismetis (approved):
I'm not going to do a deep review of this as I know we are starting by planting the flag.
Unless my one comment changes anything I think this is great!

cookiecutter.json
OKitsMikael: We don't NEED graphql if we're using React, but it does leave the API fetching layer up to the dev who generates the project. It could pretty easily be integrated with DRF on the backend.

PR #33, opened 2022-03-08
Feature/docker

paribaker:
This PR adds support for docker mentioned in this issue - 
Support is only added for local development another PR will include work for deployment

- [x] Improve README.md (move current copy to project readme instead of cookiecutter readme)
- [x] Add support for npm/node versions <---- wont do
- [ ] Add user friendly message when pipfile lock not present

### DIFF
--- a/{{cookiecutter.project_slug}}/docker-compose.yaml
+++ b/{{cookiecutter.project_slug}}/docker-compose.yaml
@@ -56,6 +56,3 @@ services:
         # this is set when calling npm run start it can be blank because at this point we have already installed the packages
         # setting this as a var is also an option but in that case using chokidar will not work as it is
         - NPM_READ_ONLY_TOKEN=''
-
-
-
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -154,7 +154,6 @@ WSGI_APPLICATION = "{{ cookiecutter.project_slug }}.wsgi.application"
 1. Heroku - we use dj_database_url to interpret Heroku's DATABASE_URL env variable.
 2. Specify DB_NAME, DB_USER, DB_PASS, and DB_HOST Directly in the env file.
 """
-
 # Update database configuration with dj_database_url
 heroku_default_db = dj_database_url.config()
 if bool(heroku_default_db):


oudeismetis:
Just tested this locally.
Other than the known issue of Rollbar being required (being handled in a separate PR) this seems to work smoothly for me.
I know we want some more capability with Docker, but I think this could be merged as-is (maybe just a couple of README tweaks).

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -0,0 +1,88 @@
...
+	docker compose up server
+
+run-d:
+	docker compose up
+
+run-client-d:
+	docker compose up client
+
+run-server-d:
+	docker compose up server
```
oudeismetis, line 64: not actually passing `-d` here

{{cookiecutter.project_slug}}/README.md
```diff
@@ -32,4 +32,40 @@ Download the ZIP file of icons that the site generates for you and paste them in
...
+- make run-d 
+
+**Run Client (-d for detached mode)**
+- make run-client
+- make run-client-d
+
+**Run Server-DB (-d for detached mode)**
+***The DB will always run with the server because it is a dependency***
+- make run-server
+- make run-server-d
```
oudeismetis, line 71: Good docs. I think a TL/DR version should be at the top of the README and stated as the preferred way to run locally.
(there is also a section of this README about Favicon's that I think should be moved to the README in the client folder)

oudeismetis: Some of these steps like the pipfile.lock step could be automatically done. Either when someone runs docker for the first time or during the initial cookiecutter step when the repo has just been created.
It's a slow process though so I could see leaving it to the dev who bootstraps. (perhaps it's own `make` step)

Steps like this though are kind of a "burn after reading" step, that are only needed during initial bootstrap and shouldn't clog up the README or scripts after that.

{{cookiecutter.project_slug}}/clients/vue3/vue.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   assetsDir: 'static',
   devServer: {
     proxy: {
+      public: process.env.VUE_APP_DEV_SERVER_PUBLIC || '0.0.0.0:8080',
```
oudeismetis, line 6: This breaks things.

whusterj:
Let me know if you'd like me to do an in-depth review of this.

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -51,7 +50,7 @@ STAFF_EMAIL='{{cookiecutter.author_name}} <{{cookiecutter.email}}>'
 #     CURRENT_PORT      The port linked to the 'primary' domain for the purpose of user-facing
 #                           links.
 #
-ALLOWED_HOSTS='localhost,127.0.0.1'
+ALLOWED_HOSTS='0.0.0.0,localhost,127.0.0.1'
```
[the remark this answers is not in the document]
oudeismetis: Pushing a change shortly for this, but the solution is to use Docker's networking.
Often something that can be setup manually, but when you use docker-compose docker with automatically create a network, name all of your containers, and add them all to that network. So if you want to reach the server, you can just put in `http://server:8000` and Docker will resolve that to the proper IP.

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -51,7 +50,7 @@ STAFF_EMAIL='{{cookiecutter.author_name}} <{{cookiecutter.email}}>'
 #     CURRENT_PORT      The port linked to the 'primary' domain for the purpose of user-facing
 #                           links.
 #
-ALLOWED_HOSTS='localhost,127.0.0.1'
+ALLOWED_HOSTS='server,localhost,127.0.0.1'
```
oudeismetis, line 53: Docker compose networking knows the backend server by the name `server`. So that needs to be added here to make Django happy.

{{cookiecutter.project_slug}}/README.md
```diff
@@ -4,32 +4,21 @@
...
-The Django app is already configured to serve favorite icons for all browsers and platforms (include, for example, apple-icons and android-icons at various sizes). By default, this icon is the vue/react logo.
-
-***Note your image must be a square otherwise a white bg will appear because the file is cropped if it is not a square go to [iloveimg.com](https://www.iloveimg.com/resize-image) and resize it.*** 
-Visit [realfavicongenerator.net](https://realfavicongenerator.net/) and upload a high resolution, square version of the image you would like to use as the favicon for this app.
-
-Download the ZIP file of icons that the site generates for you and paste them in the `client/public/static/favicons/` directory.
-
-When we run collectstatic the public folder is copied as is and enables serving of the favicons
-
-
```
oudeismetis, line 35: I think we want to make Docker the default way to run projects.
So I simplified the README here and pushed the manual steps down into the BE/FE specific READMEs.

{{cookiecutter.project_slug}}/clients/vue3/vue.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   devServer: {
     proxy: {
       '/api/': {
-        target: (process.env.VUE_APP_DEV_SERVER_BACKEND +'/api/'|| 'https://{{ cookiecutter.project_slug }}-staging.herokuapp.com') + '/api/',
+        target: (process.env.VUE_APP_DEV_SERVER_BACKEND || 'http://server:8000'),
```
oudeismetis, line 7: `.env.local` already generated with a reference to pointing at staging.
Since there are now 3 ways to find the BE (staging, local, and docker), this change means we will use the env setting if it's set. If it isn't, we'll assume docker.

{{cookiecutter.project_slug}}/scripts/deploy-on-heroku.sh
```diff
@@ -7,31 +7,31 @@ APP_NAME={{ cookiecutter.project_slug }}-staging.herokuapp.com
...
 {%- endif %}
+
 heroku config:set SECRET_KEY="$(openssl rand -base64 64)" --app $APP_NAME
 heroku config:set DEBUG="True" --app $APP_NAME
 heroku config:set CURRENT_DOMAIN="$APP_NAME.herokuapp.com" --app $APP_NAME
 heroku config:set ALLOWED_HOSTS="$APP_NAME.herokuapp.com,localhost" --app $APP_NAME
 heroku config:set NPM_CONFIG_PRODUCTION=false --app $APP_NAME
 heroku config:set ROLLBAR_ACCESS_TOKEN=$ROLLBAR_ACCSESS_TOKEN --app $APP_NAME
-fi
+
```
oudeismetis, line 34: bug

{{cookiecutter.project_slug}}/README.md
whusterj: Why are we removing the favicon instructions? Are these going somewhere else?

This is still relevant to [favicon_urls.py](https://github.com/thinknimble/tn-spa-bootstrapper/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/server/%7B%7Bcookiecutter.project_slug%7D%7D/core/favicon_urls.py), which can break production deployments if favicons are not configured correctly.

whusterj: NVM, I see it was moved to the client README

{{cookiecutter.project_slug}}/clients/vue3/README.md
```diff
@@ -37,4 +48,4 @@ See [Configuration Reference](https://cli.vuejs.org/config/).
 
 ### This use vue3
```
whusterj, line 49: ^ not part of the diff, but there is a typo here.

whusterj (approved):
Have not tested, but code looks good!

PR #41, opened 2022-03-21
fix/admin route

dan1229:


oudeismetis (approved):

whusterj:
Thanks! Just need to make sure we communicate this or document it somewhere, since it will run counter to the Django docs now.

PR #47, opened 2022-04-14
fix/professional styling

dan1229:


oudeismetis (changes requested):
Good easter eggs are better hidden 😉

whusterj:
HA

PR #50, opened 2022-04-14
Feature/caching layer

dan1229:
this one involves the caching of the related items in which they are sorted in the manner where they are properly staged for the CI we're going to implement when we Dockerize the Kubernetes API and circle back to the main priority of this iteration.


{{cookiecutter.project_slug}}/clients/vue3/src/views/Home.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="home">
-    <img alt="Project Logo" src="../assets/logo.png" />
+    <img alt="Project Logo" src="../assets/logo.jpg" />
```
oudeismetis, line 3: should be jpeg, not jpg

dan1229: >:(

PR #51, opened 2022-04-14
Fix/will i am

dan1229:
this is to increase the quality


OKitsMikael (changes requested):
Changes requested

dan1229:
> Changes requested

I think I fixed what you're talking about - I think we need to get this merged ASAP though

OKitsMikael (changes requested):
Plz change

PR #58, opened 2022-04-22
Remove yarn lock file

oudeismetis:
Lock files should be generate after a project it bootstrapped so the latest versions of libraries are used.
(In addition Dependabot was getting confused)


whusterj:
This makes sense to me - as long as we are wary of version bumps that can create breaking changes.

whusterj (approved):

PR #59, opened 2022-04-25
Require Rollbar only in Production

oudeismetis:
Rollbar is expected for Production.
Right now it errors during initial bootstrap if you don't have a token.

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -331,7 +331,7 @@ if not IN_DEV:
 #
 # Rollbar logging config
 #
-if IN_STAGING or IN_PROD:
+if IN_PROD:
     ROLLBAR_ACCESS_TOKEN = config("ROLLBAR_ACCESS_TOKEN")
 else:
     ROLLBAR_ACCESS_TOKEN = config("ROLLBAR_ACCESS_TOKEN")
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -331,8 +331,13 @@ if not IN_DEV:
 #
 # Rollbar logging config
 #
+if IN_STAGING or IN_PROD:
+    ROLLBAR_ACCESS_TOKEN = config("ROLLBAR_ACCESS_TOKEN")
+else:
+    ROLLBAR_ACCESS_TOKEN = config("ROLLBAR_ACCESS_TOKEN")
+
 ROLLBAR = {
-    "access_token": config("ROLLBAR_ACCESS_TOKEN"),
+    "access_token": ROLLBAR_ACCESS_TOKEN,
     "environment": ENVIRONMENT,
     "root": BASE_DIR,
 }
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -337,8 +337,13 @@ def _env_get_required(setting_name):
 #
 # Rollbar logging config
 #
+if IN_PROD:
+    ROLLBAR_ACCESS_TOKEN = _env_get_required("ROLLBAR_ACCESS_TOKEN")
+else:
+    ROLLBAR_ACCESS_TOKEN = os.environ.get("ROLLBAR_ACCESS_TOKEN")
```
oudeismetis, line 343: This works to unblock the initial bootstrap steps.
But I don't know if there are side effects to running Review Apps and Staging that have Rollbar set up with a bad token.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -369,23 +359,43 @@ def _env_get_required(setting_name):
...
+# Rollbar logging config
+#
+ROLLBAR_ACCESS_TOKEN = _env_get_required("ROLLBAR_ACCESS_TOKEN") if IN_PROD else os.environ.get("ROLLBAR_ACCESS_TOKEN")
+
+if ROLLBAR_ACCESS_TOKEN:
+    MIDDLEWARE += ["rollbar.contrib.django.middleware.RollbarNotifierMiddleware"]
+    ROLLBAR = {
+        "access_token": ROLLBAR_ACCESS_TOKEN,
+        "environment": ENVIRONMENT,
+        "branch": "main",
```
whusterj, line 382: Is there a reason to include the branch name here? This could be wrong if we use this for staging or review apps, right?

whusterj (approved):
Looks great! I just had one question about branch name.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: I pulled this in from another project that had it. Not sure what it does and haven't tested it.

oudeismetis: docs are a little confusing on this. On the one hand they say if you don't specify this it defaults to `master`.
But then they say `The branch field is optional, but must be set whenever the value is not master or main.`
So perhaps it's best to just remove for now and see if we run into an issue at some point with Review Apps.

oudeismetis:
https://github.com/thinknimble/tn-spa-bootstrapper/pull/60 made a big change to settings.py
Will wait on that PR to rebase and do a final test of this PR.

oudeismetis:
https://github.com/thinknimble/tn-spa-bootstrapper/pull/60 made a big change to settings.py
Will wait on that PR to rebase and do a final test of this PR.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
whusterj: Yeah, I think this is just metadata (like the environment name) that shows up in the Rollbar dashboard that you can use to sort and filter Rollbar items. If it's always 'main' then it won't be useful, so probably better to remove it here, rather than send (likely) incorrect metadata.

oudeismetis:
Rebased, tested and works locally (at least in terms of not having Rollbar on by default)
This is also the same config we used before so I'm confident it'll work.
I don't think it's worth a full test. (should be able to do that soon though)

brunogarciagonzalez (approved):
Looks good, don't see a reason to let this stagnate. I recommend we merge and if any kinks arise we tackle them then

PR #60, opened 2022-04-25
Add pytest for generated project BE validation

oudeismetis:


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/test_settings.py
```diff
@@ -0,0 +1,8 @@
+from {{ cookiecutter.project_slug }}.settings import *
+
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": os.path.join(BASE_DIR, "test_db.sqlite3"),
+    }
+}
```
whusterj, line 8: This is probably a fine default, but won't work if we use Postgres-only fields like JSONField, which is a pretty common need. Maybe we can add a comment with a sample config for postgres?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -1 +1,39 @@
...
+
+
+@pytest.mark.django_db
+def test_create_superuser():
+    superuser = User.objects.create_superuser(
+        email="test@example.com", password="password", first_name="Leslie", last_name="Burke"
+    )
+
+    assert superuser.is_superuser
+    assert superuser.last_name == "Burke"
```
whusterj, line 39: 🙌🏻

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - run: pipx install pipenv
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+          cache: 'pipenv'
+      - run: pipenv install --dev --deploy
+      - run: pipenv run pytest --mccabe --cov={{ cookiecutter.project_slug }} -vv {{ cookiecutter.project_slug }}
+      - run: pipenv run coverage report --fail-under=20
```
whusterj, line 17: I like this coverage report and the 20% threshold 👍🏻 

Do you know how easy it is to add a coverage badge to the README? It would be cool to keep this readily visible to everyone to encourage more test-writing.

whusterj (commented):
Left a couple questions/comments.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -21,6 +13,13 @@ def _env_get_required(setting_name):
 IN_STAGING = ENVIRONMENT == "staging"
 IN_PROD = ENVIRONMENT == "production"
 
+def _env_get_required(setting_name):
+    """Get the value of an environment variable and assert that it is set."""
+    setting = os.environ.get(setting_name)
+    if IN_STAGING or IN_PROD:
+        assert setting not in {None, ""}, "{0} must be defined as an environment variable.".format(setting_name)
+    return setting
```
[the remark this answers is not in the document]
oudeismetis: Yea all of the env variables cause some problems for getting lightweight pytest going. But there is a better solution to that problem.
I think you make a good call out. Especially since it undermines what the function name says it's doing, which adds complexity and confusion.
Honestly I've also disliked the use of `assert` here, but have ignored it up until now since it technically happens outside of the running app, which is where assert can be a problem.
I believe there are libraries that make managing environment variables like this easier. I've used one or two in the pass and will do a quick look to see if there is a more thoughtful solution here.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/test_settings.py
oudeismetis: Yea an in-memory lightweight DB like sqlite is generally preferred for backend tests that need a DB (Django-specific in our case). I could definitely see situations where a project would need something different. My current plan was to cross that bridge when we come to it and maybe leave it to the app owner to worry about. The bootstrapper is a good starting point, but can't cover all edge cases.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -1,31 +1,25 @@
...
 IN_STAGING = ENVIRONMENT == "staging"
 IN_PROD = ENVIRONMENT == "production"
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = _env_get_required("SECRET_KEY")
+SECRET_KEY = config("SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = _env_get_required("DEBUG") == "True"
+DEBUG = config("DEBUG", default=False, cast=bool)
```
oudeismetis, line 22: @whusterj untested but this is what the change looks like. (FYI)

whusterj: This looks great! I'd be in favor of switching everything over to this.

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
oudeismetis: I think I want to keep this as-is, but mccabe found 3 high complexity functions in the utils folder. I hadn't looked in there before. My eyes went 👀 
So I think for now leave mccabe on(?), but that means that during bootstrap tests will immediately fail until someone improves those functions.
opinions?

{{cookiecutter.project_slug}}/pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+DJANGO_SETTINGS_MODULE = {{ cookiecutter.project_slug }}.test_settings
+python_files = tests.py test_*.py *_tests.py
+addopts = -p no:warnings --strict-markers --no-migrations --reuse-db
+mccabe-complexity=5
```
oudeismetis, line 5: Alternatively we could change this to 8 or 9 to silence mccabe 😬

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: Tested and confirmed all this works (with one small typo bug found)

{{cookiecutter.project_slug}}/pytest.ini
whusterj: McCabe's "official" recommendation was to set the threshold to 10 and NIST says that in some cases it can be relaxed to 15 ([citation](https://en.wikipedia.org/wiki/Cyclomatic_complexity#cite_note-mccabe76-3))

I'm in favor of taking that recommendation and setting this to 10 - it would be great if we could be warned about functions in the 5-10 range. maybe we can use SonarCloud or something for that.

whusterj: Can you point me at the complex functions in the utils folder? It's likely I wrote those and wouldn't mind thinking about how to simplify them.

oudeismetis: for some reason I thought 5 was desired. But OK, 10 it will be for now.
I deleted the test codebase from my local so I don't have the report, but I believe it was:
`_validate_csv()`
`send_html_email()`
and there was a 3rd one, I think probably `_load_csv()`

oudeismetis: for some reason I thought 5 was desired. But OK, 10 it will be for now.
I deleted the test codebase from my local so I don't have the report, but I believe it was:
`_validate_csv()`
`send_html_email()`
and there was a 3rd one, I think probably `_load_csv()`

oudeismetis: mccabe change made

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/test_settings.py
whusterj: Cool, I'm fine with rolling with this and if it causes frequent issues, then we can revisit.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -288,13 +282,13 @@ def _env_get_required(setting_name):
...
-    AWS_SECRET_ACCESS_KEY = _env_get_required("AWS_SECRET_ACCESS_KEY")
+if config("USE_AWS_STORAGE", default=False):
+    AWS_ACCESS_KEY_ID = config("AWS_ACCESS_KEY_ID")
+    AWS_STORAGE_BUCKET_NAME = config("AWS_STORAGE_BUCKET_NAME")
+    AWS_SECRET_ACCESS_KEY = config("AWS_SECRET_ACCESS_KEY")
     AWS_S3_CUSTOM_DOMAIN = AWS_STORAGE_BUCKET_NAME + ".s3.amazonaws.com"
-    AWS_LOCATION = os.environ.get("AWS_LOCATION", "")
-    AWS_S3_REGION_NAME = _env_get_required("AWS_S3_REGION_NAME")
+    AWS_LOCATION = config("AWS_LOCATION", default="")
+    AWS_S3_REGION_NAME = config("AWS_S3_REGION_NAME")
```
whusterj, line 291: @oudeismetis - Out of curiousity, what sort of error is thrown for missing env vars?

whusterj (commented):
Quickly scrolled through the settings.py changes. Code changes look good, but I haven't run tests or anything.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: I did test this and it throws an exception, but I don't have it in my console anymore.
Can't find an example in their docs...

oudeismetis: https://github.com/henriquebastos/python-decouple/blob/master/decouple.py#L90
I believe this was what I saw in the server logs

oudeismetis:
ran it again
image

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
whusterj: Ah, just a custom exception. Cool!

whusterj:
Ha - I think I remember commenting that the CSV Parser code was too complex when I reviewed it a couple years ago.

`send_html_email` surprised me, since it's mostly a flat function. I wrote that one ages ago. Looks like it is specifically the file attachments logic that is a bit gnarly:

image

I think that's fine. It's all self-contained and the call signature is sensible and well-documented. Could probably use a unit test or two, but I can certify that exact function has worked well in prod on a few projects.

Aside, I noticed that `send_html_email` was moved to a `utils/misc.py` file. I think that may lower awareness of this function too much, when we really should try to raise awareness of it. IMO, it's really, really useful and should be our default way of sending emails - everyone expects HTML emails these days. I'd recommend at least moving it to `utils/emails.py` or maybe even into the `core` Django app so it's more readily visible. Open to suggestions.

oudeismetis:
> Ha - I think I remember commenting that the CSV Parser code was too complex when I reviewed it a couple years ago.
> 
> `send_html_email` surprised me, since it's mostly a flat function. I wrote that one ages ago. Looks like it is specifically the file attachments logic that is a bit gnarly:
> 
> image
> 
> I think that's fine. It's all self-contained and the call signature is sensible and well-documented. Could probably use a unit test or two, but I can certify that exact function has worked well in prod on a few projects.
> 
> Aside, I noticed that `send_html_email` was moved to a `utils/misc.py` file. I think that may lower awareness of this function too much, when we really should try to raise awareness of it. IMO, it's really, really useful and should be our default way of sending emails - everyone expects HTML emails these days. I'd recommend at least moving it to `utils/emails.py` or maybe even into the `core` Django app so it's more readily visible. Open to suggestions.

Yea in the mccabe score the emailing function only got a 6, so not really a bad offender anyway.
Once we are past MVP for the bootstrapped I think we could turn our eye to this. If you could, open an issue with these details.

Issue #18, opened 2021-11-14
Fix flake8 and black linting in generated projects


whusterj:
We have some pytest-cookies tests written for flake8 and black, but they are failing because the generated project is not compliant. Once we merge #12, we should quickly follow up with a PR fixing all the linting issues and then re-enable those tests.

oudeismetis:
I think this issue should remain open until Github Actions (both generated project as well as the bootstrapper) run linters to confirm no regressions here. PRs to the bootstrapper should fail for linting issues.

(Thus running black and opening a PR that touches all the files is something we should only ever have do once)

PR #66, opened 2022-04-30
Version bumps and config improvements for flake8 and flake8-isort.

whusterj:
A quick PR to improve the flake8-isort experience, based on my experience on another project.

This PR:
1. Bumps flake8 and flake8-isort to their latest versions. This is not a breaking change AFAIK.
2. Provides a configuration for flake8-isort in pyproject.toml that enables it to play nice with our preferred line length of 140 and `black` auto-formatting. I arrived at this after trial and error and trolling isort GitHub issues. Can affirm that it's working as expected for me.

### DIFF
--- a/{{cookiecutter.project_slug}}/Pipfile
+++ b/{{cookiecutter.project_slug}}/Pipfile
@@ -41,8 +41,8 @@ python_version = "3.9"
 [dev-packages]
 # Code quality
 # ------------------------------------------------------------------------------
-flake8 = "==3.9.2"  # https://github.com/PyCQA/flake8
-flake8-isort = "==4.0.0"  # https://github.com/gforcada/flake8-isort
+flake8 = "==4.0.1"  # https://github.com/PyCQA/flake8
+flake8-isort = "==4.1.1"  # https://github.com/gforcada/flake8-isort
 coverage = "==5.5"  # https://github.com/nedbat/coveragepy
 black = "==21.7b0"  # https://github.com/psf/black
 pylint-django = "==2.4.4"  # https://github.com/PyCQA/pylint-django
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -2,3 +2,11 @@
 line-length = 140
 target-version = ['py39']
 exclude = 'migrations'
+
+[tool.isort]
+py_version = 39
+line_length = 140
+include_trailing_comma = true
+use_parentheses = true
+force_grid_wrap = 0
+multi_line_output = 3


brunogarciagonzalez:
@whusterj how would I test this PR?

whusterj:
At a minimum, you should be able to run `black` and `flake8` from the command line. Steps would be:

1. Run the bootstrapper to create a new test app
2. `cd` into the new app and `pipenv install --dev`
3. `cd server`
4. `black .`
5. `flake8 .`

Ideally, you shouldn't see any changes made by `black` or problems reported by `flake8`. If you do, then we should look at fixing those in the bootstrapper. I actually haven't checked this myself. If you are seeing _conflicts_ between `black` and `flake8`, then that's a definite fail. This PR is intended to help them get along better.

Now - this particular PR is focused on `flake8-isort`, which looks for imports that are not in the correct order or format (per PEP8). To test that, I'd find a file that has a few different kinds of imports and mix them up so they aren't in the correct order. `flake8 .` should report a problem with the import order and `black .` should fix the import order automatically and make `flake8` happy.

Finally, I have my editor set up to run `black` autoformatting on save and to display `flake8` issues in real time. I really recommend doing this in your editor, and that's probably the _best_ way to test this, because you get feedback in real time. [I think my instructions on this in Notion should still be good](https://www.notion.so/thinknimble/Setting-Up-Your-Editor-for-Black-Flake8-9ad921118b57443885a91e6413bb412e), but let me know if you have trouble with this.

{{cookiecutter.project_slug}}/Pipfile
```diff
@@ -41,8 +41,8 @@ python_version = "3.9"
 [dev-packages]
 # Code quality
 # ------------------------------------------------------------------------------
-flake8 = "==3.9.2"  # https://github.com/PyCQA/flake8
-flake8-isort = "==4.0.0"  # https://github.com/gforcada/flake8-isort
+flake8 = "==4.0.1"  # https://github.com/PyCQA/flake8
+flake8-isort = "==4.1.1"  # https://github.com/gforcada/flake8-isort
```
oudeismetis, line 45: I don't think we should be version pinning by default in the Pipfile. That's what the lock file is for. Perhaps just change these to `"*"`?

oudeismetis:
When we get a chance, I'd also like to set up the bootstrapper PRs to check for linting. But that can be outside of this PR.
If no one else tests this I think I can in the next day or so.

{{cookiecutter.project_slug}}/Pipfile
whusterj: I think as a general rule it is best to at least specify major versions, and maybe even minor versions (so either `4.*` or `4.0.*` in this case). 

For well-supported and slow-moving packages like flake8, I am more inclined towards `"*"`, because I think the risk is low...

But I want to be clear that I would not do that for just any package - and even in this case there is some risk of an uncontrolled break. Not every package actually uses semver, and not all packages do semver correctly, even if they try. I have seen patch versions break builds more than once. That's even with the lock file, because in practice devs don't hesitate to destroy or update the lock. The Pipfile.lock _does_ help us trace breaking changes, but it doesn't necessarily protect us from a package being updated automatically - the Pipfile can do that, though.

paribaker:
I agree with William here, we’ve certainly had issues in the past onboarding devs because if this. Added to that I had a huge headache with black when someone was on a different version so imo even the big ones should be included.

{{cookiecutter.project_slug}}/Pipfile
oudeismetis: Yea we talked about this before and I don't want to take this PR back down that rabbit hole. 
(I have mixed opinions in my head about your argument here) 
So we can move on for now.

The one thought I'll leave here is that when a team starts a project from scratch, they usually install the latest of everything to start and then peg everything. Then the risk becomes no one ever taking the time to version bump. This risk is amplified if teams are bootstrapping new projects with a Pipfile we've provided that suffers from this same outdated issue. So you are starting a new project with an already outdated Pipfile.

Something to stew on. We can discuss later if there is a non-process way to mitigate the outdated Pipfile issue.

oudeismetis:
flake8 worked and found a bunch of issues.
Black however....
image

oudeismetis:
https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

of course technically this PR is about flake8 and not black soooo.... 😄

whusterj:
Pegging 'click' to version 8.0.4 should work.

Good case in point of why I recommend pegging versions in the Pipfile. In response to your point, if you expect project teams to bump versions when they bootstrap, you may be signing them up to detangle dependency issues like this, which could slow or halt the bootstrapping process.

oudeismetis:
Would we not want to bump black here instead?

Yea you make a good point. The priority here (as always) is a working project ASAP.
The question is if there is a good way to keep those pegged version as up to date as possible so that when they bootstrap the pegged versions are the newest best ones. (I'm imagining lots of automated tests and auto-opened PRs from something like dependabot) Ideally devs shouldn't be afraid to version bump as there should be an obvious failure during automated checks if something is bad.

whusterj:
I can look into it, but IIRC we needed this version of black for a specific reason/feature. I'll try to make a fix and test today.

oudeismetis:
This PR is related to https://github.com/thinknimble/tn-spa-bootstrapper/issues/18
(Although the scope of that issue is bigger than what this PR is doing)

oudeismetis (approved):

oudeismetis:
I didn't intend for this to happen.
I worked on a PR for black but in the process had to pull in these changes.
https://github.com/thinknimble/tn-spa-bootstrapper/pull/77/

So I think this PR will have to just be closed 😬

brunogarciagonzalez:
> I didn't intend for this to happen. I worked on a PR for black but in the process had to pull in these changes. #77
> 
> So I think this PR will have to just be closed 😬

@oudeismetis this PR includes version bumps to flake, but #77 doesn't. How to handle that? Should your PR now include that as well? Or should we merge this in and resolve conflicts in #77 ?

oudeismetis:
@brunogarciagonzalez #77 Includes that change. (you might have made your comment while I was pushing it :)

PR #67, opened 2022-05-02
Small bugs found while bootstrapping a project

oudeismetis:
I ran the bootstrapper recently to create a project and ran into a bunch of tiny issues. Willing to break out into separate PRs if this feels too big to anyone.

- [x] pytest for generated project will now run for bootstrapper PRs
- [x] bug in Github Actions for generated project
- [x] bugs is settings.py for generated project
- [x] Switch to Pipfile instead of requirements.txt for the bootstrapper
- [x] Updating an existing project with cookiecutter fails and leaves behind junk
- [x] Graphql isn't being removed if not chosen for install
- [x] USE_AWS_STORAGE bug breaks django admin


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,33 @@
...
+          cache: 'pipenv'
+      - name: Install bootstrapper dependencies
+        run: pipenv install --dev --deploy
+      - run: pipenv run cookiecutter . --no-input -f
+      - name: Pytest, McCabe, and Coverage
+        run: |
+          cd my_project
+          pipenv install --dev --deploy --skip-lock
+          pipenv run pytest --mccabe --cov=my_project -vv server/my_project
+          pipenv run coverage report --fail-under=20
```
oudeismetis, line 33: Code updates made to the generated project will now have unit tests run to validate

Pipfile
```diff
@@ -0,0 +1,27 @@
...
+pytest-cookies = "==0.6.1"
+pytest-instafail = "==0.4.2"
+PyYAML = "==5.4.1"
+PyGithub = "==1.55"
+Jinja2 = "==3.0.1"
+
+[dev-packages]
+
+[requires]
+python_version = "3.9"
```
oudeismetis, line 27: Switching to Pipfile instead of requirements.txt

.github/workflows/pytest.yml
oudeismetis: I did drop the matrix strategy (see run-pytest.yml). It was getting complicated with Pipfile and honestly we don't need it since this isn't an install-able library.

hooks/post_gen_project.py
```diff
@@ -72,177 +52,87 @@ def print_thankyou():
     )
 
 
-def remove_vue3_files():
-    shutil.rmtree(os.path.join("clients", "vue3"))
+def get_random_secret_key():
+    # Vendoring Django's get_random_secret_key() so users don't need to install it
+    chars = "abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)"
+    return "".join(secrets.choice(chars) for i in range(50))
```
oudeismetis, line 58: Instead of the complicated `generate_random_string()` function, this is a copy/paste of what Django does to generate secret keys. I did try importing Django, but that felt like a heavy dependency to require for someone just trying to generate the code first.

hooks/post_gen_project.py
```diff
@@ -72,177 +52,87 @@ def print_thankyou():
...
-
-    postgres_password = set_flag(
-        file_path,
-        "!!!POSTGRES_PASSWORD!!!",
-        value=value,
-        length=64,
-        using_digits=True,
-        using_ascii_letters=True,
-    )
-    return postgres_password
```
oudeismetis, line 169: Removed several unused functions and refactored some overly complicated logic.

hooks/post_gen_project.py
```diff
@@ -72,177 +52,87 @@ def print_thankyou():
...
-        [
-            os.path.join(
-                "..", "{{ cookiecutter.project_slug }}", "scripts", "install-reqs.sh"
-            )
-        ],
-        stdin=subprocess.PIPE,
-    )
-    shellscript.wait()
-    shellscript.stdin.close()
-
```
oudeismetis, line 231: Maybe controversial, but I don't think we should be magically installing things on the user. I've moved this to be one of the INFO statements where we tell the user the script exists if they choose to use it.

tests/test_bare.sh
```diff
@@ -1,40 +0,0 @@
...
-pipenv shell
-
-# Lint by running pre-commit on all files
-# Needs a git repo to find the project root
-git init
-git add .
-pre-commit run --show-diff-on-failure -a
-
-# run the project's tests
-pytest
```
oudeismetis, line 40: YAGNI

{{cookiecutter.project_slug}}/scripts/init-github.sh
```diff
@@ -6,6 +6,7 @@
 git init
 git add .
 git commit -m "Initial commit"
+# git remote add origin git@github.com:thinknimble/{{cookiecutter.project_slug}}.git
 gh repo create thinknimble/{{cookiecutter.project_slug}} --private -y
 git push origin main
 printf "\033[0;32mRepo https://github.com/thinknimble/{{cookiecutter.project_slug}}/\033[0m \n"
```
oudeismetis, line 12: Should improve this later, but for those who are creating the repo first on Github (and who don't want to install the `gh` utility), the comment here is the alternative approach.

oudeismetis: side comment: All these scripts assume you are pushing repos to `thinknimble`. That should not be an assumption.

hooks/post_gen_project.py
whusterj: I agree. The dev will also still have to log into the CLI tools installed by install-reqs.sh, so that could be confusing if they don't know they installed them.

whusterj (commented):
Looks good - no specific feedback from me. I'll give this a test drive soon.

brunogarciagonzalez (approved):
LGTM! Tested live w/ Melissa & Ed

Issue #63, opened 2022-04-28
Vue token management should be refactored


oudeismetis:
Per comments on https://github.com/thinknimble/tn-spa-bootstrapper/pull/61
```
"I would suggest separating token management from the user model. I think it could potentially be dangerous to include an auth token in user payloads separate from authentication workflows. It just increases the likelihood that an auth token might show up in an API endpoint where it doesn't belong.

I'd also then create store actions for obtaining a token, verifying a token, and for logout. The /verify endpoint and action are especially necessary if using tokens that can expire - like JWTs - which need to be periodically renewed.
...
You also need to consider the case where an auth-protected page is accessed from an external link or page refresh. There might be a user in the store, but you should validate their API token and redirect them before showing the page.
...
I used a convenient composable that effectively added a layer of extra code but made it easier to use the store and not have to call commit/dispatch on each store mutation/action. It also meant I didn't have to keep importing the whole store, instead I would just import the constants I would need.
One caveat is it wont work in a regular js file so in the axios you still import the store but everywhere else you can just import the composable and declare the constants you want to use."
```

whusterj:
Stale.

PR #74, opened 2022-05-05
vue3 fixed user instantiation after authentication flows

brunogarciagonzalez:
Fixed issue where instance of `User` being added to the store state was empty.

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/vue3/src/services/users/api.js
+++ b/{{cookiecutter.project_slug}}/clients/vue3/src/services/users/api.js
@@ -28,10 +28,16 @@ export default class UserAPI extends ModelAPI {
 
   login(d) {
     const data = { email: d.email.toLowerCase(), password: d.password }
-    const promise = AxiosClient.post(LOGIN_ENDPOINT, data).catch(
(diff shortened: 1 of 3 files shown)


oudeismetis:
How do you test this?

brunogarciagonzalez:
> How do you test this?

@oudeismetis I tested by running the flow locally and along with logging, used the Vue dev tools to confirm the fix that  this code prevents an empty instance of the clienside `User` model to be added to the `store.state`.

There was a bug where even with an error we were instantiating an empty user instance.
Now we have an updated vue3 `store.getters.isLoggedIn` where the token field of the user (`store.state.user.token`) must be truthy as well, so that improves the `isLoggedIn` getter as a stand-alone improvement. But this PR also fixes the bug of instantiating an empty user instance, where we would rather `store.state.user` remain `null` until a proper login/signup yields a valid instance of the clientside `user` model.

Whenever we fix #63  then `store.getters.isLoggedIn` should be updated again.

oudeismetis:
OK! I'll probably test this tomorrow

oudeismetis (approved):
tested and confirmed fix

PR #79, opened 2022-05-19
Cypress E2E tests

oudeismetis:


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: cypress-io/github-action@v2
+        with:
+          working-directory: client
+          browser: firefox
+          headless: true
+        env: 
+          NPM_CONFIG_PRODUCTION: false
+          CYPRESS_TEST_USER_EMAIL: ${{ secrets.CYPRESS_TEST_USER_EMAIL }}
+          CYPRESS_TEST_USER_PASS: ${{ secrets.CYPRESS_TEST_USER_PASS }}
+          CYPRESS_baseUrl: "https://tn-spa-bootstrapper-staging.herokuapp.com/"
```
oudeismetis, line 41: Upon this merging to staging, it'll run the Cypress tests against staging. For that to work, Github Actions needs the creds that are here and Staging needs to have this user exist (which it does). For convenience these creds can be found in the Heroku settings, which is useful should you want to point your local Cypress at staging to test something.

oudeismetis: There are a couple of challenges with this.
Right now Heroku doesn't auto deploy because of the Github linkage bug.
But even if it did, Cypress is likely to run before the deploy finishes.
In the past I've add an arbitrary 5min wait for that, but I didn't want to pollute the bootstrapper with that.
I have some ideas for a better fix.

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -141,3 +141,9 @@ SMTP_HOST='smtp.mailgun.org'
 SMTP_PORT='587'
 SMTP_VALID_TESTING_DOMAINS='thinknimble.com'
 {% endif %}
+
+# Testing (NOTE: Heroku and Github Actions will need to have matching values for some of these)
+DJANGO_SUPERUSER_PASSWORD='!!!DJANGO_SECRET_KEY!!!'
+CYPRESS_baseUrl='http://localhost:8080'
+CYPRESS_TEST_USER_PASS='!!!DJANGO_SECRET_KEY!!!'
+CYPRESS_TEST_USER_EMAIL='cypress@example.com'
```
oudeismetis, line 149: running `db_setup.sh` will inject this user with these creds into your DB so this will work.
We don't have the project specific app.json yet, but when we do it's easy to have this also run automatically for all review apps.

{{cookiecutter.project_slug}}/.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+- [ ] Todo 3
+
+
+## Deploy Notes
+
+Add any notes for changes needed when deployed. I.e. API keys, environment variables, database changes, etc.
+
+## How to test
+
+Add user steps to achieve desired functionality for this feature.
```
oudeismetis, line 19: Maybe clean this up a little, but this is the same one that is currently added to generated projects.

{{cookiecutter.project_slug}}/clients/vue3/src/views/Login.vue
```diff
@@ -15,11 +16,12 @@
         :errors="form.password.errors"
         @blur="form.password.validate()"
         type="password"
+        id="test-password"
         label="Password:"
         placeholder="Password"
       />
 
-      <button type="submit">Login</button>
+      <button id="test-submit" type="submit">Login</button>
```
oudeismetis, line 24: Not sure if there are opinions but years ago this was how we made it clear that the automated tests expected something in the HTML to exist. Not sure if there are better/cleaner ways now. Would love opinions on this.

whusterj: I think this could be a good standard, though I might suggest coming up with a custom `data-` attribute, as opposed to using IDs, so that IDs, if used, can be semantic. I'd also support just using IDs without the `test` prefix. In any case, it's going to be something the developers have to be aware of and maintain.

whusterj (commented):
This looks good so far. I'll have to test at some point. How close would you say this is?

{{cookiecutter.project_slug}}/clients/vue3/src/views/Login.vue
brunogarciagonzalez: While adding an id does work, I recommend adding data-cy attributes instead, as per https://docs.cypress.io/guides/references/best-practices#How-It-Works.

That way it is clear that these data-attributes are there for cypress to target.

bwennuh: I think the `data-*` attribute might be the best route to go here. I know I used `id`'s when doing it before, but Cypress' docs ([here](https://docs.cypress.io/guides/references/best-practices)) say the best practice is to use the `data-*` attribute so that you're using a selector that isn't also _potentially_ tied to CSS and/or JS.

oudeismetis:
Updates made.
Again, hard to confirm the tests pass here but I have a project I can share that has this testing successfully.

PR #87, opened 2022-06-15
Graphql generating settings.py formatting bugs

paribaker:
## What this does

There was a bug on the cookiecutter render, Grapehen was being rendered on the same line as the closing tag for INSTALLED_APPS causing a small bug

Additionally there was no fallback for when the index.html does not exist (aka when run build and collect static are not ran), since both react and vue can use the rendertemplate function I removed the if statements and had both use this new function. Additionally it seems the error thrown from the TemplateDoesNotExist does not get caught here so instead we check if the file exists and then render accordingly.

This should fix it!

- [ ] confirm works for vue
- [ ]  confirm works for react

### DIFF
--- a/{{cookiecutter.project_slug}}/.env.example
+++ b/{{cookiecutter.project_slug}}/.env.example
@@ -114,6 +114,49 @@ USE_AWS_STORAGE='False'
 #     AWS_LOCATION               Folder location inside of the S3 storage bucket.
 #	  AWS_S3_REGION_NAME		 Region name (ex: 'us-east-2') required in order to receive correct
 #								 query authorization mechanism parameters for private URLs.
+#      ** Due to some recent changes in aws these settings must also be made:
+#       1. create the s3 bucket
+#        2. Uncheck disable Block all public access
+#        3. Add this ACL under Bucket Policy (make sure to specify the path of the folder to allow django access so that private is still private) - Permissions
+#            ```
+#
+#            {
+#                "Version": "2012-10-17",
+#                "Statement": [
+#                    {
+#                        "Sid": "PublicReadGetObject",
+#                        "Effect": "Allow",
+#                        "Principal": "*",
+#                        "Action": "s3:GetObject",
+#                        "Resource": "arn:aws:s3:::<bucket-name>/<folder>/static/*"
+#                    }
+#                ]
+#            }
+#            ```
+#            and in CORS - Permissions
+#            ```
+#                [
+#                    {
+#                        "AllowedHeaders": [
+#                            "Authorization"
+#                        ],
+#                        "AllowedMethods": [
+#                            "GET"
+#                        ],
+#                        "AllowedOrigins": [
+#                            "*"
+#                        ],
+#                        "ExposeHeaders": [
+#                            "x-amz-server-side-encryption",
+#                            "x-amz-request-id",
+#                            "x-amz-id-2"
+#                        ],
+#                        "MaxAgeSeconds": 3000
+#                    }
+#                ]
+#            ```
+#
+#
 #
 AWS_ACCESS_KEY_ID=''
 AWS_SECRET_ACCESS_KEY=''
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
@@ -1,5 +1,54 @@
 from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin
 
 from .models import User
 
-admin.site.register(User)
+
+class CustomUserAdmin(UserAdmin):
+    model = User
+
+    fieldsets = (
+        (
+            None,
+            {
+                "fields": (
+                    "email",
+                    "first_name",
+                    "last_name",
+                    "password",
+                    "last_login",
+                    "is_staff",
+                )
+            },
+        ),
+    )
+
+    add_fieldsets = (
+        (
+            None,
+            {
+                "classes": ("wide",),
+                "fields": ("email", "password1", "password2"),
+            },
+        ),
+    )
+
+    list_display = ("is_active", "email", "first_name", "last_name")
+
+    list_display_links = (
+        "is_active",
+        "email",
+        "first_name",
+        "last_name",
+    )
+
+    search_fields = (
+        "email",
+        "first_name",
+        "last_name",
+    )
+
+    ordering = []
+
+
+admin.site.register(User, CustomUserAdmin)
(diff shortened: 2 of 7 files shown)


oudeismetis (commented):
Please fix linting errors

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -294,16 +301,14 @@
...
-    STATIC_URL = f"https://{aws_s3_domain}/static/"
-    MEDIA_URL = f"https://{aws_s3_domain}/media/"
 
-#
-# STATIC
-# ------------------------
-STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
+    STATIC_URL = f"https://{aws_s3_domain}/{STATIC_FILES_LOCATION}/"
+    MEDIA_URL = f"https://{aws_s3_domain}/{PRIVATE_MEDIAFILES_LOCATION}/"
+
```
oudeismetis, line 311: Not your change, but it's a little odd that all these vars are in all CAPS except for `aws_s3_domain`

oudeismetis: In fact there is some weirdness above that we could clean up here.
`aws_s3_domain = AWS_S3_CUSTOM_DOMAIN or f"{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com"`
This will ALWAYS use `AWS_S3_CUSTOM_DOMAIN` since that seems to always be set

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -22,22 +22,19 @@
...
-    except TemplateDoesNotExist:
-        return render(request, "core/index-placeholder.html")
-{% endif %}
+        index_path = os.path.join(settings.BASE_DIR.replace('server', 'client'), 'dist/index.html')
+        if not os.path.exists(index_path):
+            raise FileNotFoundError
+        return TemplateResponse(request, "index.html")
+    except FileNotFoundError:
+        return TemplateResponse(request, "core/index-placeholder.html")
+
```
oudeismetis, line 37: I don't think you need the try/except here.
I think instead you can just return the placeholder template inside of `if not os.path.exists(index_path):`

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -114,6 +114,49 @@ USE_AWS_STORAGE='False'
...
+#                        ],
+#                        "ExposeHeaders": [
+#                            "x-amz-server-side-encryption",
+#                            "x-amz-request-id",
+#                            "x-amz-id-2"
+#                        ],
+#                        "MaxAgeSeconds": 3000
+#                    }
+#                ]
+#            ```
```
whusterj, line 157: This is nice documentation. I do wonder whether this is the most appropriate place for it, but I think it can live here for now, because it's helpful.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/storages.py
```diff
@@ -0,0 +1,13 @@
...
+
+class StaticStorage(S3Boto3Storage):
+    location = settings.STATIC_FILES_LOCATION
+
+
+class PrivateMediaStorage(S3Boto3Storage):
+    location = settings.PRIVATE_MEDIAFILES_LOCATION
+    default_acl = "private"
+    file_overwrite = False
+    custom_domain = False
```
whusterj, line 13: Ah, I was wondering if this was still needed. 👍🏻

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -211,36 +188,47 @@
...
+                "django.contrib.auth.context_processors.auth",
+                "django.contrib.messages.context_processors.messages",
+            ]
+        },
+    }
+]
 
-STATICFILES_FINDERS = [
-    "django.contrib.staticfiles.finders.FileSystemFinder",
-    "django.contrib.staticfiles.finders.AppDirectoriesFinder",
```
whusterj, line 243: ^ This change might be unexpected to people familiar with Django. I know I'm accustomed to being able to add static files in the app directories and have those auto-detected. I don't feel strongly about this, but is there a reason that this _should_ be removed?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/templates/core/index-placeholder.html
```diff
@@ -8,8 +8,23 @@
...
+      <h1> Hi, Welcome to the Django Portal </h1>
+      You are seeing this page for one of two reasons
+      <ul>
+      <li>You have not built your client side </li>
+      <li>Or you have built your client side and haven't collected your static files</li>
+      <ul>
+      <i><strong>To fix this run:</strong> </i>
+      <ul>
+      <li><code> npm run build || yarn run build </code> </li>
+      <li> <code> python server/manage.py collectstatic --no-input</code> </li>
```
whusterj, line 23: I think I'm fine with this change, and it's definitely helpful, but bear in mind that if the build process fails on Heroku or whatever hosting environment, then there's a chance this will end up being a public-facing message.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
```diff
@@ -1,5 +1,54 @@
...
+    search_fields = (
+        "email",
+        "first_name",
+        "last_name",
+    )
+
+    ordering = []
+
+
+admin.site.register(User, CustomUserAdmin)
```
whusterj, line 54: 👍🏻 Looks good!

whusterj (commented):
Looks good overall! I had a couple thoughts.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -211,36 +188,47 @@
...
+{% if cookiecutter.client_app == "Vue3" %}
+CLIENT_BUILD_DIR = "../client/dist/"
+CLIENT_STATIC_FILES = "../client/dist/static"
 {% elif cookiecutter.client_app == 'React' %}
-STATICFILES_DIRS = [
-    os.path.join(STATIC_ROOT, "static"),
-]
-{% endif -%}
-{% endif -%}
+CLIENT_STATIC_FILES = "../build/static"
```
OKitsMikael, line 205:
```suggestion
CLIENT_STATIC_FILES = "../client/build/static"
```

OKitsMikael (changes requested):
@paribaker left a couple of comments/fixes for the React side of things, otherwise looks good. I generated the project from this branch and tested that it serves React correctly after collecting static files, so it should be good to go once the changes are made 👍

PR #89, opened 2022-06-22
Cypress config Setup step

oudeismetis:


{{cookiecutter.project_slug}}/.github/workflows/cypress.yml
```diff
@@ -1,18 +1,31 @@
...
+      - name: Set URL for PR
+        if: success() && steps.findPr.outputs.number
+        run: echo "BASE_URL=https://{{ cookiecutter.project_slug }}-pr-${PR}.herokuapp.com/" >> $GITHUB_ENV
+        env:
+          PR: {{ "${{ steps.findPr.outputs.pr }}" }}
+      - name: Raise for error
+        if: env.BASE_URL == ''
+        run: exit 1
+    outputs:
+      BASE_URL: {{ "${{ env.BASE_URL }}" }}
```
oudeismetis, line 21: This is the real change with this PR.
This has been used on a couple of projects now and is a cleaner way to figure out which URL to hit for Cypress.
The first time a PR is created the `gh-find-current-pr` step always seems to fail. Problem is we would then cache that failure so all subsequent runs would also fail. Thats fixed here as we fail the step if we don't get a proper url.

.github/workflows/cypress.yml
```diff
@@ -1,41 +1,81 @@
...
+on: push
 jobs:
+  Setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: jwalton/gh-find-current-pr@v1
+        id: findPr
+      - name: Set URL for Staging
+        if: github.ref_name == 'main'
+        run: echo "BASE_URL=https://tn-spa-bootstrapper-stage.herokuapp.com/" >> $GITHUB_ENV
```
bwennuh, line 11: Should be
```
run: echo "BASE_URL=https://tn-spa-bootstrapper-staging.herokuapp.com/" >> $GITHUB_ENV
```
(vs `-stage`)

oudeismetis: good catch

bwennuh (approved):
Looks good to me!

Issue #43, opened 2022-03-24
Add default 404 page for the Vue app


oudeismetis:
When generating a Vue/Django app with the bootstrapper, missing URL routes aren't handled correctly.
We should handle 404s properly with an acceptable default page for newly generated projects.

PR #92, opened 2022-07-11
Feature/ Vue 404 Page

bwennuh:
## What this does

Issue #43: Add default 404 page for Vue applications for any URL paths that are not currently defined in the project's Vue router.

## Checklist
- [x] Create new `PageNotFound.vue` view component to display if page is not found
- [x] Add new path routing in `/router/index.js` to handle paths that do not exist

## How to test

Either open the review app or create/run a bootstrapper app with this branch (`feature/vue-404-page`) and navigate to a URL that is not currently pathed with the Vue router (`/test`, for example) and you should be able to view the `PageNotFound` view, which has the 404 error code at the top, as well as some informational text.

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
+++ b/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
@@ -1,0 +1,18 @@
+<template>
+  <div class="page-not-found">
+    <h1><span class="error">404</span> Page Not Found</h1>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'PageNotFound',
+  components: {},
+}
+</script>
+
+<style scoped lang="scss">
+.error {
+  color: red;
+}
+</style>
--- a/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
+++ b/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
@@ -12,7 +12,6 @@
 <script>
 export default {
   name: 'PageNotFound',
-  components: {},
 }
 </script>
 
--- a/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
+++ b/{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
@@ -1,6 +1,11 @@
 <template>
   <div class="page-not-found">
-    <h1><span class="error">404</span> Page Not Found</h1>
+    <h1 class="page-not-found__error">404</h1>
+    <img alt="Project Logo" src="../assets/logo.png" />
+    <h2 class="page-not-found__header">Oops!<br />We couldn't find the page you're looking for.</h2>
+    <p class="page-not-found__route">
+      Click <router-link to="/">here</router-link> to return to the home page.
+    </p>
   </div>
 </template>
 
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div class="page-not-found">
+    <h1 class="page-not-found__error">404</h1>
+    <img alt="Project Logo" src="../assets/logo.png" />
```
bwennuh, line 4: **Question:** is it worth adding the logo here? I think it adds a little more to the page, but then folks will have to remember to replace the logo image with the [client] project specific one and not sure if then it'd be easier to just not have an image here.

If we did want to keep the logo here, however, we could always add a reminder/note to the Vue `README.md` file.

{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
```diff
@@ -0,0 +1,31 @@
...
+    <h2 class="page-not-found__header">Oops!<br />We couldn't find the page you're looking for.</h2>
+    <p class="page-not-found__route">
+      Click <router-link to="/">here</router-link> to return to the home page.
+    </p>
```
bwennuh, line 8: **Info:** Just went with some basic text to indicate that the page the user has landed on does not exist (as well as giving a clickable/focusable redirect link for users), but this can be updated to something else should additional/different verbiage be preferred.

{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
```diff
@@ -0,0 +1,30 @@
...
+  name: 'PageNotFound',
+}
+</script>
+
+<style scoped lang="scss">
+.page-not-found {
+  &__error {
+    margin: unset;
+    font-weight: 800;
+    font-size: 64px;
```
bwennuh, line 23: **Info:** Added an arbitrary larger font size for the "404" text just to display it larger on the page to make it more obvious to visual users that they've landed on a not-found page. Font size can be increased or decreased if needed.

bwennuh: **Question:** Do we want the "404" error code to be displayed on the page at all? And if so, do we want the text to have a different color and/or style to further separate it from the rest of the text?

{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
oudeismetis: This page existing is good enough for now.
Eventually all pages should be at least half-decently designed with things like logos in a good location.
There was a block of comments in the README for how to swap out the favicons and logos. That got removed but will get added back soon. Thats another step devs should be doing shortly after bootstrapping.

sooo...perhaps having the logo here is good as it reminds the devs they still need to do that step?

{{cookiecutter.project_slug}}/clients/vue3/src/views/PageNotFound.vue
oudeismetis: We've been holding off on too much polish on these pages, largely because we wanted to discuss the design system stuff first. I think once a decision is made there an effort can be made to create a consistent default look and feel to all the default pages.

oudeismetis (approved):

PR #98, opened 2022-07-22
Vue Password Reset steps are broken

oudeismetis:
## What this does
Learned about this issue while using the bootstrapper for a project.

## Checklist
- [x] Fix "Send me a password reset email" page erring on submit
- [x] Fix "Create new password" page not saving a new password
- [x] Test on Review App
- [x] Write test if possible


oudeismetis:
image

Testing evidence that the backend at least tries to send a password reset email. (main bug for this PR)

oudeismetis:
This should mostly work, but needs someone to help it across the finish line (I don't have the time at the moment)

paribaker:
@oudeismetis is this still a thing we need?

oudeismetis:
It is. I'm actually implementing emails on a project now so I have a reason to test this.

oudeismetis:
This is close to done. Everything seems to work (albeit with copy/paste of url from the logs)

Will wait for https://github.com/thinknimble/tn-spa-bootstrapper/pull/162 to be merged before doing a final full test of this PR.

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
```diff
@@ -29,5 +29,7 @@ jobs:
           touch .env
           cat .env.example > .env
           cat .env
+      - run: mkdir -p client/dist/static
+      - run: pipenv run python server/manage.py collectstatic
```
oudeismetis, line 33: I don't love adding this, but we now have a test that expects a static image to exist.
I thought about mocking it in some way, but ultimately it's probably better to have it test with real things instead of mocked things.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from {{ cookiecutter.project_slug }}.core.models import User
+
+
+@pytest.fixture
+def user(db):
+    return User.objects.create(email="user@example.com", password="1234", first_name="test", last_name="user")
```
oudeismetis, line 8: The beginnings of a conftest. There are other fixtures we have that should get moved here, but I'll do that in a more testing focused PR.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -63,3 +65,18 @@ def test_user_can_login(test_user):
...
+    test_user.save()
+    context = test_user.reset_password_context()
+    password_reset_url = f"/api/password/reset/confirm/{ context['user'].id }/{ context['token'] }/"
+    response = client.post(password_reset_url, data={"password": "new_password"}, format="json")
+    assert response.status_code == 200
+
+    # New Password should now work for authentication
+    serializer = UserLoginSerializer(data={"email": test_user.email, "password": "new_password"})
+    serializer.is_valid()
+    assert authenticate(**serializer.validated_data)
```
oudeismetis, line 82: This test hits the API endpoint to do the password reset, and then confirms that the new password now works.

Interesting thing here is how it actually hits the API instead of calling the function directly.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/emails.py
```diff
@@ -5,7 +5,14 @@
...
 from django.template.loader import render_to_string
-from premailer import transform
+from premailer import transform as inline_css
+
+
+def get_html_body(template, context):
+    # two steps:
+    # Render the django jinja template to string with passed context
+    # Use premailer to inline CSS styles for weird email clients
+    return inline_css(render_to_string(template, context))
```
oudeismetis, line 15: Refactored this for two reasons:

- Easier to write tests
- `transform()` is vague
- `render_to_string()` is vague

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/emails.py
```diff
@@ -37,34 +44,19 @@ def send_html_email(subject, template, send_from, send_to, context={}, bcc_email
     #         if domain not in valid_domains:
     #             return
 
-    assert isinstance(send_to, (list, tuple, str)), "send_to must be an instance of list, tuple, or str"
+    if not isinstance(send_to, (list, tuple, str)):
+        raise Exception("send_to must be an instance of list, tuple, or str")
```
oudeismetis, line 48: You shouldn't really use `assert` in real code as it doesn't do proper checking. It should be used in tests only.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/tests.py
```diff
@@ -30,3 +31,9 @@ def test_get_site_url_negative(settings, custom_settings):
         settings.__setattr__(key, custom_settings[key])
     with pytest.raises(Exception):
         get_site_url()
+
+
+def test_password_reset_email_link(user):
+    context = user.reset_password_context()
+    html_body = get_html_body("registration/password_reset.html", context)
+    assert f"{ context['site_url'] }/password/reset/confirm/{ context['user'].id }/{ context['token'] }" in html_body
```
oudeismetis, line 39: This tests the other half of password reset. Making sure the URL that is generated and emailed matches our expectations.

With these two tests in place, password reset shouldn't easily break for projects. It's one of those features that no one notices is broken until a real user can't change their password.

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
paribaker: why not?

{{cookiecutter.project_slug}}/clients/web/vue3/src/router/index.js
```diff
@@ -15,28 +15,31 @@ const routes = [
     // this generates a separate chunk (about.[hash].js) for this route
     // which is lazy-loaded when the route is visited.
     component: () => import(/* webpackChunkName: "login" */ '../views/Login.vue'),
+    beforeEnter: requireNoAuth,
   },
```
paribaker, line 19: I think we should merge these two into one chunk rather than two they are both very small

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/RequestPasswordReset.vue
```diff
@@ -1,20 +1,43 @@
...
+        <button
+          :class="{ 'btn--disabled': !form.isValid }"
+          :disabled="!form.isValid"
+          type="submit">
+          Request Password Reset
+        </button>
+      </form>
+    </div>
+  </template>
+  <template v-else>
```
paribaker, line 23: We should pull these out into constants for translation services in a near future

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
```diff
@@ -9,16 +9,30 @@
         label="New Password:"
         placeholder="New password"
       />
+      <InputField
+        v-model:value="form.passwordConfirmation.value"
```
paribaker, line 13: FYI with the new tn-forms you can pass in labels and placeholders by default

paribaker: Hmm will fix this we already have form level validators!

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
```diff
@@ -29,11 +43,27 @@ export default {
   },
   setup() {
     const form = ref(new PasswordResetForm())
+    let matchingPasswordsError = ref(false)
+    const store = useStore()
     const route = useRoute()
+    const router = useRouter()
```
paribaker, line 49: Can remove this!

paribaker (approved):

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
```diff
@@ -29,11 +43,27 @@ export default {
...
+      () => [form.value.password.value, form.value.passwordConfirmation.value],
+      () => {
+        if (form.value.password.value && form.value.passwordConfirmation.value
+            && form.value.password.value !== form.value.passwordConfirmation.value){
+          matchingPasswordsError.value = 'Passwords must match'
+        }else{
+          matchingPasswordsError.value = false
+        }
+      }
+    )
```
paribaker, line 61: @oudeismetis since you added this I can leave it here or ask Tomas to give it a shot!
We have a pre-build Validator for this [here](https://github.com/thinknimble/tn-forms/blob/2698ab22409678984034b533c51a879af2e11c78/src/validators.ts#L103) you can see an example of it in the `tn-forms` readme.md in the JS section [here]("https://github.com/thinknimble/tn-forms"). 

Check out the comment in the `forms.js`

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/forms.js
```diff
@@ -33,4 +33,10 @@ export class PasswordResetForm extends Form {
       new MinLengthValidator({ minLength: 8, message: 'Minimum Length of 8 required' }),
     ],
   })
+  static passwordConfirmation = new FormField({
+    validators: [
+      new RequiredValidator(),
+      new MinLengthValidator({ minLength: 8, message: 'Minimum Length of 8 required' }),
+    ],
+  })
 }
```
paribaker, line 42: For form level validators there is a special key (eventually need to make it a symbol)
```
export class PasswordResetForm extend Form{
 ....fields....
    // add this 
   static dynamicFormValidators = {
    confirmPassword: [new MustMatchValidator({ matcher: 'password' })],
  }
  }
}
```

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
oudeismetis: something about needing to collectstatic in order to run unit tests feels wrong to me. In this case I think it's a good thing.

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/forms.js
oudeismetis: This (and the tn-forms docs) confuse me a little here. Given that this is password confirmation, shouldn't I just drop the two existing validators (redundant) and just add the MustMatchValidator? Or is the whole dynamicFormValidators step somehow special?

oudeismetis: I tried the idea I stated above and things are now broken. It always says the passwords don't match even when they do.

{{cookiecutter.project_slug}}/clients/web/vue3/src/router/index.js
paribaker: The chunks here are saying that each one of these files is it's own chunk. The reason we chunk them is for lazy loading in this case we are saying we should lazy load them all but really that just generates a bunch of small file chunks that don't really affect size/loading time

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/forms.js
paribaker: yea it has to be set up with the dynamicFormValidators instead this is the case for all FormLevel validators. The reason is that we cannot get the reference to the value until the form is created (i.e the same pointer in memory). The dynamic form validator object is set after we create the form and feeds the validator the form field.

paribaker: but you are right you can just drop the two existing ones if you want I add them just as an additional measure to remind the user to fill them in.

paribaker:
@oudeismetis left you a comment, the gist of it is the MustMatchValidator is a form level validator and needs to be added with the dynamicFormValidators key OR if you wanted to post form creation with the helper method. !

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/forms.js
paribaker: @oudeismetis did you want me to complete this for you?

oudeismetis: yes please

paribaker (approved):

Issue #88, opened 2022-06-16
Run Cypress after Heroku Deploy


oudeismetis:
Currently the Cypress tests run when the other tests run. The problem is that Heroku finishes deploying the new changes after the tests run, so the Cypress results are actually for the previous code push.

Better solutions:

1. A Heroku post-deploy script makes an API call to Github (https://api.github.com/repos/{repo-name}/dispatches)
2. The Github Action for Cypress is changes to only trigger on `repository_dispatch`

https://niteo.co/blog/staging-like-its-2020

oudeismetis:
A final fix for this was implemented in https://github.com/thinknimble/tn-spa-bootstrapper/pull/150

PR #99, opened 2022-07-25
Cypress release trigger

oudeismetis:
Better way to run Cypress for Review Apps

Addresses https://github.com/thinknimble/tn-spa-bootstrapper/issues/88

- [x] Github Action turns on Heroku maintenance mode temporarily so Cypress defaults to failing
- [x] Heroku deployments call a release script
- [x] Release script calls Github API to trigger a rerun of the failing Cypress run
- [x] 2nd run of Cypress passes and is now guaranteed to be testing the most current version of deployed code
- [x] Document secret generation for devs

TODO:
- [x] Implement feature and/or fix bug
- [x] Add and/or fix tests
- [x] Review App looks good
- [x] Code Review
- [x] During merge...remove `NPM_PRIVATE_TOKEN` from Github and Heroku settings
- [x] During merge...generate new `GITHUB_TOKEN` and add it to the pipeline and staging settings in Heroku
- [x] During merge...generate new `HEROKU_API_KEY` and replace the old one in Github settings

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -66,27 +66,17 @@ jobs:
   Firefox:
     needs: Setup
     runs-on: ubuntu-latest
-    container:
-      image: cypress/browsers:node16.14.2-slim-chrome103-ff102  # https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
-      options: --user 1001
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - run: |
-          # Cypress slim container image doesn't come with pipx
-          apt update -qq
-          apt install -y python3-venv python3-pip
-          # apt-get install python3-venv python3-pip
-          # python3 -m pip install pipx
-          # python3 -m pipx ensurepath
-          python3 -m pip install --user pipenv
+      - run: pipx install pipenv
       - uses: actions/setup-python@v2
         with:
           python-version: '3.9'
           cache: 'pipenv'
       - name: Install bootstrapper dependencies
-        run: python3 -m pipenv install --dev --deploy
-      - run: python3 -m pipenv run cookiecutter . --no-input -f
+        run: pipenv install --dev --deploy
+      - run: pipenv run cookiecutter . --no-input -f
       - name: Install frontend dependencies
         env:
           NPM_CONFIG_PRODUCTION: false
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -77,9 +77,9 @@ jobs:
           apt update -qq
           apt install -y python3-venv python3-pip
           # apt-get install python3-venv python3-pip
-          python3 -m pip install pipx
-          python3 -m pipx ensurepath
-          pipx install pipenv
+          # python3 -m pip install pipx
+          # python3 -m pipx ensurepath
+          python3 -m pip install pipenv
       - uses: actions/setup-python@v2
         with:
           python-version: '3.9'
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -77,11 +77,6 @@ jobs:
       - name: Install bootstrapper dependencies
         run: pipenv install --dev --deploy
       - run: pipenv run cookiecutter . --no-input -f
-      - name: Install frontend dependencies
-        env:
-          NPM_CONFIG_PRODUCTION: false
-        working-directory: ./my_project/client
-        run: npm install
   FirefoxRun:
     needs: Firefox
     runs-on: ubuntu-latest
@@ -89,6 +84,11 @@ jobs:
       image: cypress/browsers:node16.14.2-slim-chrome103-ff102  # https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
       options: --user 1001
     steps:
+      - name: Install frontend dependencies
+        env:
+          NPM_CONFIG_PRODUCTION: false
+        working-directory: ./my_project/client
+        run: npm install
       - name: Run against ${{ needs.Setup.outputs.BASE_URL }}
         uses: cypress-io/github-action@v4
         with:
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -49,7 +49,7 @@ jobs:
         working-directory: ./my_project/client
         run: npm install
       - name: Run against ${{ needs.Setup.outputs.BASE_URL }}
-        uses: cypress-io/github-action@v2
+        uses: cypress-io/github-action@v4
         with:
           working-directory: my_project/client
           browser: chrome
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -78,6 +78,7 @@ jobs:
           apt install -y python3-pip
           # apt-get install python3-venv python3-pip
           python3 -m pip install --user pipx
+          python3 -m pipx ensurepath
           pipx install pipenv
       - uses: actions/setup-python@v2
         with:
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -67,7 +67,7 @@ jobs:
     needs: Setup
     runs-on: ubuntu-latest
     container:
-      image: cypress/browsers:cypress/browsers:node16.14.2-slim-chrome103-ff102  # https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
+      image: cypress/browsers:node16.14.2-slim-chrome103-ff102  # https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
       options: --user root
     steps:
       - name: Checkout
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -9,12 +9,10 @@ jobs:
         id: findPr
       - name: Set URL for Staging
         if: github.ref_name == 'main'
-        run: |
-          echo "BASE_URL=https://tn-spa-bootstrapper-staging.herokuapp.com/" >> $GITHUB_ENV
+        run: echo "BASE_URL=https://tn-spa-bootstrapper-staging.herokuapp.com/" >> $GITHUB_ENV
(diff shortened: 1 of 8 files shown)


oudeismetis:
Has been tested on a live project. Adding bug fixes and code review workoffs items here soon.

.github/workflows/cypress.yml
```diff
@@ -1,5 +1,11 @@
 name: Cypress Tests
-on: push
+on:
+  workflow_call:
+    secrets:
+      CYPRESS_TEST_USER_EMAIL:
+        required: true
+      CYPRESS_TEST_USER_PASS:
+        required: true
```
oudeismetis, line 8: `workflow_call` means it's called my other workflows (instead of on code push).
As a result, it doesn't have its own env or secrets so those have to be passed explicitly.

.github/workflows/cypress.yml
```diff
@@ -17,12 +23,15 @@ jobs:
...
-    steps:
+      - name: Raise for maintenance mode
+        run: |
+          html_body=$(curl ${{ env.BASE_URL }})
+          if echo $html_body | grep -q "Offline for Maintenance";
+          then
+            exit 1
+          else
+            echo "NOT in maintenance. Likely Heroku trigger after successful deploy";
+          fi
```
oudeismetis, line 34: Test the URL to see if the app is not up yet (first commit) or if the app is in maintenance mode (waiting on deploy to finish)

.github/workflows/cypress.yml
```diff
@@ -33,36 +42,51 @@ jobs:
       - name: Install bootstrapper dependencies
         run: pipenv install --dev --deploy
       - run: pipenv run cookiecutter . --no-input -f
+      - uses: actions/upload-artifact@v3
+        with:
+          name: my_project
+          path: my_project/
+          retention-days: 1
+    outputs:
+      BASE_URL: ${{ env.BASE_URL }}
```
oudeismetis, line 51: `upload-artifacts` allows us to do all the work we needed to do for cookiecutter only once. The later workflows can then do things like run inside a docker container and not worry about pip missing or not having permissions to install things (problem I ran into)

.github/workflows/cypress.yml
```diff
@@ -33,36 +42,51 @@ jobs:
...
+  Chrome:
+    needs: Setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: actions/download-artifact@v3
+        with:
+          name: my_project
+          path: my_project/
```
oudeismetis, line 61: Only step needed to get the results of building code out during `Setup`

.github/workflows/cypress.yml
```diff
@@ -33,36 +42,51 @@ jobs:
...
       - name: Install frontend dependencies
         env:
           NPM_CONFIG_PRODUCTION: false
         working-directory: ./my_project/client
         run: npm install
```
oudeismetis, line 66: But for some reason `npm install` during `Setup` wouldn't work. Didn't bother digging into why.

.github/workflows/cypress.yml
```diff
@@ -33,36 +42,51 @@ jobs:
...
       - name: Run against ${{ needs.Setup.outputs.BASE_URL }}
-        uses: cypress-io/github-action@v2
+        uses: cypress-io/github-action@v4
         with:
           working-directory: my_project/client
           browser: chrome
-          headless: true
```
oudeismetis, line 46: no longer a thing. It's headless by default. Use `headed` to actually have a browser available.

.github/workflows/cypress.yml
```diff
@@ -33,36 +42,51 @@ jobs:
...
           CYPRESS_TEST_USER_EMAIL: ${{ secrets.CYPRESS_TEST_USER_EMAIL }}
           CYPRESS_TEST_USER_PASS: ${{ secrets.CYPRESS_TEST_USER_PASS }}
           CYPRESS_baseUrl: ${{ needs.Setup.outputs.BASE_URL }}
   Firefox:
-    runs-on: ubuntu-latest
     needs: Setup
+    runs-on: ubuntu-latest
+    container:
+      image: cypress/browsers:node16.14.2-slim-chrome103-ff102  # https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
+      options: --user 1001
```
oudeismetis, line 82: Firefox should now be more stable when it runs. (lots of questions/confusion/complaints online about running Firefox with Cypress. Hopefully this fix holds for a while)

.github/workflows/heroku_deploy.yml
```diff
@@ -0,0 +1,44 @@
...
+    if: ${{ always() }}
+    needs: [Setup, Cypress]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Turn Maintenance Mode Off for ${{ needs.Setup.outputs.APP_NAME }} 
+        run: |
+          echo "Turning off maintenance so Heroku triggering rerun of Cypress will succeed";
+          heroku maintenance:off --app=${{ needs.Setup.outputs.APP_NAME }}
+        env:
+          HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
```
oudeismetis, line 44: On every push of code we'll now immediately:

- Put the app in maintenance mode
- Run Cypress expecting it to fail fast (otherwise it'll be hidden on the PR and it won't exist to be re-run later)
- Turn maintenance mode off

{{cookiecutter.project_slug}}/Procfile
```diff
@@ -14,6 +14,6 @@ web: gunicorn {{ cookiecutter.project_slug }}.wsgi --chdir=server --log-file -
 #   3. Up-to-date build numbers shown in app (ideally auto-generated at build-and-deploy time)
 #
 # Comment the line below to disable migrations automatically after a Heroku push.
-release: python server/manage.py makemigrations --noinput && python server/manage.py migrate --noinput
+release: python server/manage.py makemigrations --noinput && python server/manage.py migrate --noinput && ./scripts/release.sh
```
oudeismetis, line 17: When Heroku is finished deploying it'll call this new shell script

{{cookiecutter.project_slug}}/README.md
```diff
@@ -4,6 +4,10 @@
 
 ## Setup
 
+### Github & Heroku
+1. [Generate an auth token for Heroku](https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-authorizations-create) and add it to the [repo secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) so Github Actions can reach Heroku. `heroku authorizations:create -d "Github Actions" -s write-protected`
+1. [Generate an auth token for Github](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#authenticating-with-the-api) and add it as [an environment variable](https://devcenter.heroku.com/articles/config-vars) so Heroku can trigger Github Actions
```
oudeismetis, line 9: Heroku auth token is easy to generate and lives forever. Adding it to Github settings is safe as well.
The Github auth token is currently a personal auth token of mine (Ed) with an expiration in a week or so. That could be a long-lived token, but it also required authing the token with JumpCloud. Not sure how long that lives for. PLUS, you then add that token to Heroku and all devs can see it, so not as safe.

I did look into a solution for that. Github Actions DOES generate a temporary app token while the workflows are running and I was able to pass that to Heroku using the Heroku CLI (same spot where we turn maintenance mode on). Problem is that the token gets expired when the workflow finishes. Seems like the only way around that is to keep the workflow alive....so then we're back to sleeping for 5min.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/management/commands/trigger_cypress.py
```diff
@@ -0,0 +1,39 @@
...
+    def handle(self, *args, **kwargs):
+        self.headers["Authorization"] = f"Bearer {config('GITHUB_TOKEN')}"
+        org = "thinknimble"
+        repo = "tn-spa-bootstrapper"
+        branch = config("HEROKU_BRANCH")
+        self.github_root = f"https://api.github.com/repos/{org}/{repo}"
+
+        workflow_id = self.find_cypress_workflow_id(branch)
+        job_id = self.find_job_id_for_cypress_step(workflow_id)
+        self.rerun_job(job_id)
```
oudeismetis, line 39: A little complicated/convoluted, but basically we call Github's API to find the most recent Cypress run on this PR and tell it to run again.
At this point deployment is finished so rerunning will not only pass, but Cypress will be testing the most recently version of the code.

.github/workflows/cypress.yml
brunogarciagonzalez: If it was a react app then it could be due to that `Apollo` peer dependencies error that is also behind the ESLint tests failing

oudeismetis: It was on a Vue app. I think the error was something about the lock file missing. Something not getting carried over during the upload/download artifact step. I'm not really worried, though I like the idea of doing all the build steps during setup only.

oudeismetis: I understand this better now. Makes perfect sense now that I think about.
Storing artifacts via `actions/upload-artifact` is only for files and folders. For things like installs via apt/python/npm there is a separate `caches` github action (`actions/cache`) which is also available inside actions like `actions/setup-python` (which we use). That is the proper way to install something once, and then reuse it between jobs or even separate workflow runs.
You can in fact install something when running the `main` branch and it'll use that cache by default on PRs.
artifacts are less powerful and are attached to a specific workflow run. So they can't be reused more widely.

brunogarciagonzalez (approved):
🚀 🚀 🎸 ✅

oudeismetis:
Future me here 👋 
This is not a good solution.
There is a much better and simpler way to accomplish this.
See https://github.com/thinknimble/tn-spa-bootstrapper/pull/150

PR #101, opened 2022-08-02
Boilerplate favicon and meta tags for client apps

whusterj:
## What this does

Adds some smart defaults and standards for setting up favicons, a preview image, and the Facebook and Twitter meta tags. This setup works well for 

## Checklist
- [x] Add favicon tags to vue3 and react index.html files
- [x] Add meta tags to vue3 and react index.html
- [x] Verify that `settings.py` staticfiles settings will work with both clients
- [x] Add default favicons to both clients (?)
- [ ] Add default preview image to both clients (?)

## Deploy Notes

These meta tags are good defaults, but require some customization by engineers when deploying. As a standard, we use favicon-generator.org with a SQUARE image to generator the set of icon files and the manifest.json file referenced by index.html and by favicon_urls.py.

Beyond that, I recommend customizing the preview image, description tags, and possibly the URL tags.

## How to test

TODO: Need to run some tests myself.


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -215,25 +217,33 @@
 #
 # Static files (CSS, JavaScript, Images)
 #
-{% if cookiecutter.client_app != "React" -%}
-STATIC_ROOT = os.path.join(BASE_DIR, "static")
-MEDIA_ROOT = os.path.join(BASE_DIR, "media-files")
-{% else %}
-STATIC_ROOT = os.path.join(BASE_DIR, "..", "client", "build")
-MEDIA_ROOT = os.path.join(BASE_DIR, "..", "client", "build", "static")
-{% endif -%}
```
whusterj, line 224: ^ Since these are the 'target' directories where static or media files will be stored, there's no reason for them to differ between Vue and React.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -215,25 +217,33 @@
...
+# Django will look for client-side build files in this directory
 STATICFILES_DIRS = [
     os.path.join(BASE_DIR, "../client/dist/static"),
 ]
 {% elif cookiecutter.client_app == 'React' %}
+# Django will look for client-side build files in this directory
 STATICFILES_DIRS = [
-    os.path.join(STATIC_ROOT, "static"),
+    os.path.join(BASE_DIR, "..", "client", "build"),
+    os.path.join(BASE_DIR, "..", "client", "build", "static"),
```
whusterj, line 246: ^ @OKitsMikael for your review: These are the settings that worked on our recent client project for React. Note that we're importing both the outer `client/build` folder and the inner `client/build/static`. This means files in `client/build/static` are copied to the `STATIC_ROOT` twice. On the first pass they are copied to `static/static` and the second time they are copied to `static/`.

This is a hack. I had to do this, because Django Whitenoise was otherwise not able to resolve the path to some font files referenced in the CSS. Those font files were in `/static/static/media/`, but it was looking in `/static/media/`, so copying the `client/static/` folder to the top level of `STATIC_ROOT` resolved the issue.

Confusing, I know. Let know if that makes sense.

whusterj (commented):
I think all the pieces are in place so this is ready for review and testing.

oudeismetis:
love that you replaced those gross placeholder images on the vue side of things

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
OKitsMikael: @whusterj the bit about copying files to `STATIC_ROOT` twice seems a little odd to me, my understanding was that Django would look in this list of directories for static files once `collectstatic` runs and dumps the static files into `STATIC_ROOT`, thus merging any assets on the server side into the same folder that's used to hold the client-side static files.

I also see that on L228, you've changed `STATIC_ROOT`, so the files will no longer be put into `client/build/static` but instead `/static`. Not sure what else this change will affect.

I think we could maybe play with the `STATIC_URL` and `MEDIA_URL` so that we don't copy the same files into multiple directories if that's indeed what's going on. I'll generate a test project, and load it into a review app to make sure that fonts, images, and favicons all work properly.

oudeismetis: @OKitsMikael If you change this file: https://github.com/thinknimble/tn-spa-bootstrapper/blob/main/buildpack-run.sh
You should be able to have it generate a react app for the review app on this PR to ease testing of these changes. (I've been thinking about how to automate this going forward)

whusterj: Your description of `collectstatic` sounds correct, and the merging of client and server-side files is effectively what happens with our config.

Maybe a more thorough breakdown will clarify things: `STATICFILES_FINDERS` lists the classes used to find static files. First up is `FileSystemFinder`, which looks in the `STATICFILES_DIRS` and copies those over to `STATIC_ROOT`, followed by `AppDirectoriesFinder`, which grabs the contents of the `static/` directory in each Django app.

1. So first, it copies the contents of `client/build` to `STATIC_ROOT`
2. Second, it copies the contents of `client/build/static` to `STATIC_ROOT`
3. Third, it copies the contents of the `static` directory from each Django app to `STATIC_ROOT`

So the contents of `client/build/static` ends up existing in two places: `STATIC_ROOT/static` after step one and `STATIC_ROOT/` after step 2. Again, I only did this because of an error thrown by Django Whitenoise when running `collectstatic`. It said it could not resolve a font name that was referenced in a CSS file and existed in `client/build/static/media/`. It did not like the file being in `$STATIC_ROOT/static/media/`, but it was fine with `$STATIC_ROOT/media/`, and the second step above accomplishes this.

I think this may actually be a Whitenoise bug, but haven't had time to really look at it. Ideally, we should only need step 1 from the above list.

>I also see that on L228, you've changed STATIC_ROOT, so the files will no longer be put into client/build/static but instead /static. Not sure what else this change will affect.

It shouldn't affect anything else. The choice of directory name and location is mostly arbitrary. The only two limitations I can think of are that you don't want to overwrite existing directories, and that the process that runs Django must have permission to access it. So AFAIK, this won't have any side effects.

Also, to be clear, it's not set to the root directory `/static`. It's `$BASE_DIR/../static/`, where `$BASE_DIR` is the `server/` directory. So this puts the `static` directory at the top level of the _project_ (that is, "above" `client` and `server`). On Heroku, for instance, this ends up resolving to `/app/static`.

>I think we could maybe play with the STATIC_URL and MEDIA_URL so that we don't copy the same files into multiple directories if that's indeed what's going on

I don't think changing these settings would make a difference. These are just the URLs or path names Django uses to generate links to static and media files for use on the front end.

One other note, `MEDIA_ROOT` should never have been set to `client/build/static`, since it [is intended for user-uploaded files](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-MEDIA_ROOT), not static files. So it should be kept entirely separate, and in prod probably not kept on the filesystem at all.

OKitsMikael: @whusterj Thanks for the thorough explanation! This gives me a much clearer understanding of the process. There shouldn't be many files in `client/build/static`, so I can see how copying them twice isn't ideal, but it's not that much of an issue.

Thanks for fixing `MEDIA_ROOT`. We definitely hit the issue with user-uploaded media files getting deleted on deploy (which we resolved by using a Heroku addon for user-uploaded images).

OKitsMikael (approved):
Looks good!

oudeismetis (approved):

oudeismetis:
Cypress firefox issue. Merging anyway

PR #106, opened 2022-08-22
React Improvements

OKitsMikael:
## What this does

Adds some updates to aid React development.

## Checklist
- [x] Set null token on logout function in AuthProvider
- [ ] Logout unit tests
- [ ] Namespace auth token

## Deploy Notes

N/a

## How to test

1. Verify that React unit tests pass.

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/react/src/utils/auth.tsx
+++ b/{{cookiecutter.project_slug}}/clients/react/src/utils/auth.tsx
@@ -15,13 +15,8 @@ import { PUBLIC_ROUTES } from "./routes";
 export const AuthContext = React.createContext({
   token: null,
   updateToken: (token: any) => {}, // set default signature to expect function
+  logout: () => {}
 });
-
-export function logout() {
-  client.clearStore();
-  localStorage.setItem("auth-token", null);
-  localStorage.setItem("exp", null);
-}
 
 export function AuthProvider({ children }: { children: any }) {
   let navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/clients/react/src/utils/auth.tsx
```diff
@@ -32,7 +27,14 @@ export function AuthProvider({ children }: { children: any }) {
...
   }
-  const [token, setToken] = useState(storedToken);
+  const [token, setToken] = useState<string | null>(storedToken);
+
+  function logout() {
+    client.clearStore();
+    localStorage.setItem("auth-token", JSON.stringify(null));
+    localStorage.setItem("exp", JSON.stringify(null));
+    setToken(null)
+  }
```
whusterj, line 37: I had TypeScript issues here on a client project. I think `localStorage.setItem` expects a value with a `String` type. `JSON.stringify` might be solving that here. IIRC There were one or two other places where these values were set to `null` and causing type errors.

Any reason not to use `localStorage.removeItem` here? That gets around the type issue, and I think it makes sense that if we don't have a value it should be cleared from localStorage, unless the 'in-between' state somehow signifies something.

OKitsMikael: @whusterj `localStorage.removeItem` is the right approach. I think I'm going to close this in favor of https://github.com/thinknimble/tn-spa-bootstrapper/pull/107, since that covers this change as well as the refreshToken implementation that wasn't quite working properly in the auth module.

PR #108, opened 2022-08-29
AWS Config Settings

jdephil:
## What this does

Fixed configs for AWS s3 storage

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## Deploy Notes

Add any notes for changes needed when deployed. I.e. API keys, environment variables, database changes, etc.

## How to test

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/handlers.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/handlers.py
@@ -2,6 +2,10 @@ from django.conf import settings
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
+class StaticStorage(S3Boto3Storage):
+    location = settings.STATIC_FILES_LOCATION
+
+
 class PrivateMediaStorage(S3Boto3Storage):
     location = settings.PRIVATE_MEDIAFILES_LOCATION
     default_acl = "private"
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -298,7 +298,6 @@ STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 PRIVATE_MEDIAFILES_LOCATION = ""
 # Django Storages configuration
 if config("USE_AWS_STORAGE", cast=bool, default=False):
-    print("configging correct")
     AWS_ACCESS_KEY_ID = config("AWS_ACCESS_KEY_ID")
     AWS_STORAGE_BUCKET_NAME = config("AWS_STORAGE_BUCKET_NAME")
     AWS_SECRET_ACCESS_KEY = config("AWS_SECRET_ACCESS_KEY")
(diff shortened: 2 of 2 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -242,7 +242,7 @@
 # Django will look for client-side build files in this directory
 STATICFILES_DIRS = [
     os.path.join(BASE_DIR, "..", "client", "build"),
-    os.path.join(BASE_DIR, "..", "client", "build", "static"),
+    os.path.join(BASE_DIR, "..", "client", "build", "media-files"),
```
whusterj, line 245: The media-files directory should _not_ be included in `STATICFILES_DIRS`. This change will also break the front end app!

"Static" files are the mostly-unchanging files served by the web application. Things like: CSS, JS, fonts, and hard-coded images.

"Media" files are dynamic files, typically uploaded by users.

They serve different purposes. They can both be stored on S3, but their configuration needs to be kept separate!

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/handlers.py
```diff
@@ -2,6 +2,10 @@
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
+class StaticStorage(S3Boto3Storage):
+    location = settings.STATIC_FILES_LOCATION
```
whusterj, line 6: FYI: I don't think we have ever used S3 to host static files. On Heroku, we've always used Django Whitenoise.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -303,17 +308,15 @@
 
     aws_s3_domain = AWS_S3_CUSTOM_DOMAIN or f"{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com"
     # Default file storage is private
+
     PRIVATE_MEDIAFILES_LOCATION = AWS_LOCATION + "/media"
-    DEFAULT_FILE_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.PrivateMediaStorage"
-    STATICFILES_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.StaticRootS3Boto3Storage"
+    STATIC_FILES_LOCATION = AWS_LOCATION + "/static"
+    DEFAULT_FILE_STORAGE = "miles_for_military.core.handlers.PrivateMediaStorage"
+#   #STATICFILES_STORAGE = "miles_for_military.core.handlers.StaticStorage"
```
whusterj, line 315: Need to replace `miles_for_military` here with `{{ cookiecutter.project_slug }}`.

As mentioned above, we've never used S3 for staticfiles, so we can probably remove `STATICFILES_STORAGE` here, unless there's some reason we want to make S3 Static files work.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
paribaker: Sorry I had asked for this change I may have not made it clear @jdephil I thought it was line 229 that was set incorrectly!

paribaker:
this is not needed at the moment anymore all static assets are served on heroku, only user assets are saved on aws

PR #113, opened 2022-09-13
React ESLint: Decprecated apollo-link breaking things

### LINKED ISSUE #112: React client: Remove apollo-link deprecated package
apollo-link packages have been deprecated by the apollo team : https://github.com/apollographql/apollo-link 

Apparently they are already including the same functionalities in their v3 (which is already part of our react client)
https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#apollo-link-

Seems like the only change required is to replace the imports to point to:
```ts
import {createHttpLink} from '@apollo/client/link/http'
import {setContext} from '@apollo/client/link/context'
```

lakardion:
Closes #112

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/react/package.json
+++ b/{{cookiecutter.project_slug}}/clients/react/package.json
@@ -15,8 +15,6 @@
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "@types/react-router-dom": "^5.3.2",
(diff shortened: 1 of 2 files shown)


oudeismetis (approved):
This is good fix for the ESLint issue we've seen recently, thanks!
Heroku deploys are failing for some (unrelated) reason right now. Going to ignore Cypress and merge as-is.

PR #115, opened 2022-09-21
Generate and test a React app if React files are touched

oudeismetis:
## What this does

Right now, when you open a PR for the bootstrapper, it'll only create a Vue app and deploy that to Heroku. When Github Actions run, they also assume Vue and will run linters and automated tests against a Vue app.
This charge detects what code you touched the most (React or Vue) and will generate/test an app that matches.

## Checklist
- [x] Create Vue and React specific templates that can be passed to cookiecutter when it runs
- [x] Create a script that determines what files, when compared to `main`, have changed in this PR. Count and return if more react or more vue files
- [x] Cookiecutter picks a template based on the results of the above script
- [x] Set a default `BUILDPACK_RUN` value
- [x] Test what happens if the first deploy is react, but then a vue app is deployed (node cache issue maybe)

## How to test

Inspect Heroku Review App to confirm it's a React app. Inspect Github actions to confirm they ran against a React app instead of Vue.

## Issues/Caveats

Heroku will build immediately with the initial value of `BUILDPACK_RUN`, which uses React to deploy. The value of that Heroku var will change if most of the files touched are Vue files, but that will not effect the build until the following run. Cypress will likely fail as a result. So push another commit or manually trigger a build in Heroku for that situation.

This isn't very obvious as a dev is likely to be scratching their head wondering why Cypress failed.


.github/workflows/cypress.yml
```diff
@@ -41,7 +41,9 @@ jobs:
           cache: 'pipenv'
       - name: Install bootstrapper dependencies
         run: pipenv install --dev --deploy
-      - run: pipenv run cookiecutter . --no-input -f
+      - run: |
+          config_file=$(./scripts/vue_or_react.sh)
+          pipenv run cookiecutter . --config-file $config_file --no-input -f
```
oudeismetis, line 46: There is now a function which will detect what type of app is being worked on (details below)

.github/workflows/heroku_deploy.yml
```diff
@@ -42,3 +43,17 @@ jobs:
...
+          if echo "$current_config" | grep -q "config_file"; then
+            echo "Keeping current config of $config_file"
+          else
+            echo "Switching Heroku to build with $config_file instead"
+            heroku plugins:install heroku-builds
+            heroku builds:cache:purge --app=${{ needs.Setup.outputs.APP_NAME }} --confirm=${{ needs.Setup.outputs.APP_NAME }}
+            heroku config:set BUILDPACK_RUN="pipenv run cookiecutter . --config-file $config_file --no-input -f" --app=${{ needs.Setup.outputs.APP_NAME }}
+          fi
+        env:
+          HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
```
oudeismetis, line 59: `BUILDPACK_RUN` will default to React.
Should we detect a change in the code the dev is working on (i.e. Vue instead of React) we'll make a couple of calls here to tell Heroku to build a Vue app instead. This is done last as we currently have no way to sneak this in before Heroku starts building. So Heroku will build the wrong app and Cypress will likely fail. But the next push of code will work as expected.

.github/workflows/linting.yml
```diff
@@ -12,13 +15,18 @@ jobs:
           cache: 'pipenv'
       - name: Install bootstrapper dependencies
         run: pipenv install --dev --deploy
-      - run: pipenv run cookiecutter . --no-input -f
+      - run: |
+          config_file="cookiecutter/${{ matrix.fe_type }}_template.yaml"
+          pipenv run cookiecutter . --config-file $config_file --no-input -f
```
oudeismetis, line 20: Most of our checks should run every time against both React and Vue versions of a build.
The matrix strategy makes this easy to do.
Should we have flake8 linting errors where a change formats fine for a React app but break for Vue, we'll now catch those on the PR.

app.json
```diff
@@ -4,6 +4,9 @@
         "ALLOWED_HOSTS": {
             "value": ".herokuapp.com"
         },
+        "BUILDPACK_RUN": {
+            "value": "pipenv run cookiecutter . --config-file cookiecutter/react_template.yaml --no-input -f"
+        },
```
oudeismetis, line 9: Default for `BUILDPACK_RUN` is react.
As a reminder, If you look lower in this file you'll see the `https://github.com/weibeld/heroku-buildpack-run.git` buildpack. Heroku buildpacks let you perform different tasks during the build. This one lets us run an arbitrary command. In this case it's cookiecutter.

cookiecutter/vue_template.yaml
```diff
@@ -0,0 +1,3 @@
+default_context:
+  client_app: "Vue3"
+  use_graphql: "n"
```
oudeismetis, line 3: cookiecutter templates. Used for automation here, but you can also use them for local testing so you can skip the questionnaire.

scripts/vue_or_react.sh
```diff
@@ -0,0 +1,14 @@
...
+ts_count=$(git diff --name-only origin/main -- | grep "\.ts$" | wc -l)
+tsx_count=$(git diff --name-only origin/main -- | grep "\.tsx$" | wc -l)
+react_count=$(($ts_count + $tsx_count))
+vue_count=$(git diff --name-only origin/main -- | grep "\.vue$" | wc -l)
+
+if [ $react_count -gt $vue_count ]; then
+  echo "cookiecutter/react_template.yaml"
+else
+  echo "cookiecutter/vue_template.yaml"
+fi
```
oudeismetis, line 14: Script to determine Vue or React.
Basically we look at the list of files changed when comparing this branch to main.
We then count the number of files based on extensions. (ts/tsx vs vue)
The larger count wins.
Default for 1st build is React

{{cookiecutter.project_slug}}/Pipfile
```diff
@@ -31,7 +31,7 @@ drf-nested-routers = "==0.93.3"
 django-anymail = "==8.4"  # https://github.com/anymail/django-anymail
 rollbar = "==0.16.2"
 {%- if cookiecutter.use_graphql == "y" %}
-graphene_django = "==2.15.0"
+graphene_django = "*"
```
oudeismetis, line 34: build bug that was prevent react apps from bootstrapping

{{cookiecutter.project_slug}}/clients/react/src/pages/LogIn.tsx
```diff
@@ -102,7 +104,7 @@ export default function LogIn() {
               <></>
             )}
           </FormControl>
-          <Button onClick={handleLogin}>Sign In</Button>
+          <Button data-cy="submit" onClick={handleLogin}>Sign In</Button>
```
oudeismetis, line 107: React apps now have Cypress tests.
I was hoping to skip this but having Cypress was kinda critical for testing this PR.
Also it would have been harder to have one bootstrapped app using Cypress and the other not having it.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -16,9 +21,7 @@
 
 from .models import User
 from .permissions import CreateOnlyPermissions
-{% if cookiecutter.use_graphql == 'n' -%}from .serializers import UserLoginSerializer, UserRegistrationSerializer, UserSerializer{% else %}
-from .serializers import UserLoginSerializer, UserSerializer
-{% endif %}
+from .serializers import UserLoginSerializer, UserRegistrationSerializer, UserSerializer
```
oudeismetis, line 24: Found and fixed a bunch of linting issues while working this PR.
Because of this PR, this should be the LAST time we play the tennis match of linter errors.

scripts/vue_or_react.sh
paribaker: this may start to fail when we move Vue over to TS (although vue wont use tsx) why not target the specific client folder?

paribaker (approved):

PR #117, opened 2022-10-17
index.html template fallback for React apps

whusterj:
## What this does

React issue reported by Jennifer: favicon and metatag link calls in `index.html` of the React app create a server error: `TemplateDoesNotExist`.

The problem is twofold:
1. We don't really need those metatags in development, but since they are in index.html, the browser is requesting them, and the React dev server is proxying those requests to the backend for some reason.
2. Our Django app is configured to look for the client-side `index.html` template in `client/dist/`, but in development - as intended - this folder does not exist yet, b/c we have not built the frontend for production. The `favicon_urls` config is disabled in development, so it falls back to serving `index.html`, which fails with a `TemplateDoesNotExist error.

I think the quick fix is to add this fall back to `index-placeholder.html`, which was a strategy we were using with Vue. We should also look into configuring React to NOT proxy these calls for favicons to the backend OR add template tags to disable the meta tags altogether in dev.

## How to test

Bootstrap a React app w/ GraphQL enabled and run the client-side dev server.

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
@@ -31,10 +31,7 @@ from .serializers import UserLoginSerializer, UserSerializer
 @ensure_csrf_cookie
 @never_cache
 def index(request):
-    try:
-        return TemplateResponse(request, "index.html")
-    except TemplateDoesNotExist:
-        return TemplateResponse(request, "core/index-placeholder.html")
+    return TemplateResponse(request, ["index.html", "core/index-placeholder.html"])
 {% elif cookiecutter.client_app.lower() == 'None' %}
 def index(request):
     return redirect(to="/docs/swagger/")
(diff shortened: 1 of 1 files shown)


jdephil (dismissed):
thank you for fixing!

jdephil (approved):

OKitsMikael:
@whusterj I believe the request proxy is happening due to the `"proxy": "http://localhost:8000",` line in `client/package.json` I believe removing it should stop the dev server from proxying the favicon and meta tag links. This setup was intended to make it easy to make requests to the API without having to explicitly allow `localhost:3000` via CORS (which is already done 😅), but since we are relying on GraphQL and ApolloClient, it should be safe to take out without hurting anything.

whusterj:
@OKitsMikael - Thank you for clarifying! That confirms my suspicion that it was dev proxy similar to what we have on the Vue side. In Vue, we sorta solve this by only proxying requests to `/api/`. I'm honestly not sure what ends up happening to these metatag requests. I think they don't get proxied and the Vue dev server routes to 'not found' or ignores them.

PR #123, opened 2022-10-19
added login to react

jdephil:
## What this does

Adds a basic Log In page to React

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/react/src/pages/LogIn.tsx
+++ b/{{cookiecutter.project_slug}}/clients/react/src/pages/LogIn.tsx
@@ -1,0 +1,129 @@
+import React, { useState, useContext } from "react"
+import {
+  Heading,
+  Input,
+  Text,
+  Grid,
+  GridItem,
+  Button,
+  VStack,
+  Link,
+  Image,
+  Show,
+  Hide,
+  FormControl,
+  FormErrorMessage,
+} from "@chakra-ui/react"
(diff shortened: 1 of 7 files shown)


{{cookiecutter.project_slug}}/clients/react/src/pages/LogIn.tsx
```diff
@@ -0,0 +1,129 @@
...
+          </Show>
+          <Heading mt={5}>PORTAL LOG IN</Heading>
+          <Text>Enter your login credentials below</Text>
+          <FormControl isInvalid={error}>
+            <Input
+              placeholder="Email"
+              onChange={(e) => {
+                setEmail(e.target.value)
+              }}
+              my={5}
```
brunogarciagonzalez, line 81: For these to be fully-controlled components, do we not want a prop of
```jsx
<Input value={email} />
```

and same for the password input?

brunogarciagonzalez (commented):
Tested client side locally and looks good!
Also the GQL query looks good -- tested via dev tools network tab.

One comment/question regarding controlled components.

{{cookiecutter.project_slug}}/clients/react/src/pages/LogIn.tsx
jdephil: good catch, fixed!

brunogarciagonzalez (approved):

PR #125, opened 2022-10-21
added prettier ignore file

paribaker:
## What this does

Adds a prettier ignore file to ignore django template files from autoformatting


OKitsMikael (approved):
👍

oudeismetis (approved):

PR #127, opened 2022-10-24
React sign up

jdephil:
## What this does

Adds sign up for React/graphql


{{cookiecutter.project_slug}}/clients/react/src/pages/SignUp.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const { register, handleSubmit } = useForm<FormValues>()
+
+  const [logIn] = useMutation(LOG_IN, {
+    onCompleted: (data: any) => {
+      localStorage.setItem("auth-token", data.tokenAuth.token)
+      updateToken(data.tokenAuth.token)
+      navigate("/home")
+    },
+    onError: (error: any) => {},
```
OKitsMikael, line 41: @jdephil can we display this error in the UI also?

{{cookiecutter.project_slug}}/clients/react/src/pages/SignUp.tsx
```diff
@@ -0,0 +1,145 @@
...
+            id="confirmPassword"
+            placeholder="Confirm Password"
+            type="password"
+            onChange={(e) => {
+              setConfirmPassword(e.target.value)
+            }}
+          />
+          <FormErrorMessage>Passwords do not match</FormErrorMessage>
+        </FormControl>
+        <button
```
OKitsMikael, line 122: @jdephil does this button need `type="submit"`

OKitsMikael (commented):
Nice work! Left a couple questions

{{cookiecutter.project_slug}}/clients/react/src/pages/SignUp.tsx
jdephil: it doesn't need it, maybe because it's inside the form but i can put it in if that's better

OKitsMikael: No need to if it works! Was just curious

{{cookiecutter.project_slug}}/clients/react/src/pages/SignUp.tsx
OKitsMikael: @jdephil Hmm good question. There shouldn't be an issue between signing up and logging in here, but if there were, I think it's best practice to display the error even if we redirect the user to log in

OKitsMikael (approved):
Looks good! 👍

PR #129, opened 2022-11-02
Added app.json and fixed bugs with Python3.10, Cypress, Swagger docs

oudeismetis:
## What this does

Fixing multiple issues found while bootstrapping an app recently.

## Checklist
- [x] `CYPRESS_TEST_USER_EMAIL` isn't a secret and should just be hardcoded
- [x] pytest coverage report not in the `.gitignore` file
- [x] pytest/flake8 broke after Python 3.10 upgrade
- [x] Clearer docs about new requirements for Heroku/Github API keys
- [x] Added an `app.json`
- [x] `NPM_READ_ONLY_TOKEN` no longer a requirement
- [x] Upgrade Node/NPM
- [x] Improve errors for if/when Heroku fails to trigger Cypress to run
- [x] Remove hardcoded reference to this bootstrapper repo
- [x] Improve URL routes, default display, and permissions for API Swagger docs
- [x] More sensible defaults in `settings.py`
- [x] Don't run Cypress after deploying to production


paribaker (approved):

{{cookiecutter.project_slug}}/app.json
```diff
@@ -0,0 +1,49 @@
...
+            "value": "True"
+        },
+        "DISABLE_COLLECTSTATIC": {
+            "value": 1
+        },
+        "ENVIRONMENT": {
+            "value": "development"
+        },
+        "PROJECT_PATH": {
+            "value": "modat_api"
```
OKitsMikael, line 17: @oudeismetis ruh roh. This should be generic

{{cookiecutter.project_slug}}/app.json
```diff
@@ -0,0 +1,49 @@
...
+    },
+    "addons": [
+        "heroku-postgresql:standard-0",
+        "papertrail:Choklad"
+    ],
+    "environments": {
+        "review": {
+            "addons": [
+                "heroku-postgresql:hobby-dev",
+                "papertrail:Choklad"
```
OKitsMikael, line 34: @oudeismetis what's the use case for papertrail in review apps?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -35,9 +35,9 @@
 #
 # Domain Configuration
 #
-CURRENT_DOMAIN = config("CURRENT_DOMAIN")
+HEROKU_APP_NAME = config("HEROKU_APP_NAME", default="{{ cookiecutter.project_slug }}-staging")
+CURRENT_DOMAIN = config("CURRENT_DOMAIN", default=f"{HEROKU_APP_NAME}.herokuapp.com")
```
OKitsMikael, line 39: @oudeismetis I don't think we should supply a default for `HEROKU_APP_NAME`, that one is populated by Heroku automatically on review apps

OKitsMikael (commented):
A few comments, otherwise looks good

{{cookiecutter.project_slug}}/app.json
oudeismetis: It's free and the best way to view logs from the app when you are debugging

{{cookiecutter.project_slug}}/app.json
oudeismetis: ugh

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: For review apps it is but not for staging. This adds to the list of things that have to be manually set up by devs. We are trying in the next weeks/months to automate at much as the bootstrapping process as possible.

Getting there means making some quick decisions with the bootstrapper that we might change later. (i.e. a script that creates pipelines for you in Heroku and automatically sets all the env vars like this one so we don't need a default here)

{{cookiecutter.project_slug}}/app.json
OKitsMikael: I haven't added it to my workflow, is it easier than `heroku logs -t`? That's what I usually use

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
OKitsMikael: What is `HEROKU_APP_NAME` used for in staging?

I've found that there are some extra settings to disable when using a freshly bootstrapped app that aren't necessary at the beginning. We should consider writing some add-on scripts for those things so that we don't have unused/unneeded settings that get in the way of a first deploy, and aren't lingering around when we want to hand off a repo to a client dev. That way the decision can be made later and we'll still save time by having it automated.

oudeismetis: yea I agree. A lot of the Heroku stuff in particular I'm planning to pull out completely. We've talked about "Heroku in 20min". I'm going to start working on that soon but it might not be in this repo.
The main reason I added this here for the moment is that I just made a list of all the tedious manual steps one has to do to bootstrap, and some like these are easy to have a sensible default for to cut down on the manual config one has to do.
I also realize that I'm moving faster on PRs like these. Pari and I are trying to make significant progress on the bootstrapper before the end of the year and thus are trying to avoid PRs that sit open for weeks/months on end. So some issues will sneak in, but they should all get resolved in short order (hopefully).

PR #131, opened 2022-11-15
Improve Swagger default settings

oudeismetis:
## What this does

A bunch of small quality of life improvements, mostly with how we generate the Swagger backend API docs. See list below. 

## Checklist
- [x] Swagger docs public by default
- [x] Generate API keys on user creation instead of login
- [x] Swagger "Try it Out" feature works and uses the users own API key in the example `curl`
- [x] Update Heroku config to use `mini` for Postgres on Review Apps
- [x] Remove hardcoded references

## Deploy Notes

n/a

## How to test

View Swagger docs in Heroku.
Use the "Try it Out" Swagger feature and observe that the example `curl` request uses an auth token and not session auth.

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/vue3/public/index.html
+++ b/{{cookiecutter.project_slug}}/clients/vue3/public/index.html
@@ -40,10 +40,10 @@
     <!-- Facebook Unfurling Meta Tags -->
     <meta property="og:type" content="website" />
     <meta property="og:url" content="%PUBLIC_URL%" />
-    <meta property="og:title" content="Squirrel" />
+    <meta property="og:title" content="{{ cookiecutter.project_name }}" />
     <meta property="og:description" content="Add a description for {{ cookiecutter.project_name }} in index.html!" />
     <meta property="og:image" content="%PUBLIC_URL%/static/meta-preview.png" />
-    
+
     <!-- Twitter Unfurling Meta Tags -->
     <meta name="twitter:card" content="summary_large_image">
     <!-- TODO: Add the app domain here once live. -->
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/client/templates/drf-yasg/swagger-ui.html
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/client/templates/drf-yasg/swagger-ui.html
@@ -1,0 +1,18 @@
+{% extends "drf-yasg/swagger-ui.html" %}
+
+{% block extra_scripts %}
+    <script>
+        let authorization = {
+            Token: {
+                name: "Token",
+                schema: {
+                    type: "apiKey",
+                    "in": "header",
+                },
+                value: "Token {{ user.auth_token.key | safe }}"
+            }
+        }
+
+        localStorage.setItem(KEY_AUTH, JSON.stringify(Immutable.fromJS(authorization)))
+    </script>
+{% endblock %}
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/urls.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/urls.py
@@ -38,8 +38,8 @@ schema_view = get_schema_view(
         contact=openapi.Contact(email="support@{{ cookiecutter.project_slug }}.com"),
         license=openapi.License(name="BSD License"),
     ),
-    public=True,
-    permission_classes=[permissions.AllowAny],
+    public=False,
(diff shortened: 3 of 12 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/client/templates/drf-yasg/swagger-ui.html
```diff
@@ -0,0 +1,18 @@
...
+                name: "Token",
+                schema: {
+                    type: "apiKey",
+                    "in": "header",
+                },
+                value: "Token {{ user.auth_token.key | safe }}"
+            }
+        }
+
+        localStorage.setItem(KEY_AUTH, JSON.stringify(Immutable.fromJS(authorization)))
```
whusterj, line 16: Where does `KEY_AUTH` come from? Guessing it's a global defined somewhere in `swagger-ui.html`...

{{cookiecutter.project_slug}}/clients/vue3/public/index.html
```diff
@@ -40,10 +40,10 @@
     <!-- Facebook Unfurling Meta Tags -->
     <meta property="og:type" content="website" />
     <meta property="og:url" content="%PUBLIC_URL%" />
-    <meta property="og:title" content="Squirrel" />
+    <meta property="og:title" content="{{ cookiecutter.project_name }}" />
```
whusterj, line 43: Nice 👍🏻

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/urls.py
```diff
@@ -38,8 +38,8 @@
         contact=openapi.Contact(email="support@{{ cookiecutter.project_slug }}.com"),
         license=openapi.License(name="BSD License"),
     ),
-    public=False,
-    permission_classes=[permissions.IsAuthenticated],
+    public=True,
+    permission_classes=[permissions.AllowAny],
```
whusterj, line 42: I can understand the convenience of this, but I have doubts that this is a smart 'production-ready' default. I think it would be very easy for a dev to overlook this and accidentally expose API docs to the world when they didn't mean to.

Perhaps it can be open when `DEBUG=True` and hidden behind auth when `DEBUG=False` ? That would be consistent with Django's behavior of showing a detailed stack trace and other dev tools when `DEBUG=True`.

whusterj (commented):
Looks good - have a suggestion about handling permissions on the schema_view for Swagger.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/client/templates/drf-yasg/swagger-ui.html
oudeismetis: yea this is the logic Brenna figured out for Savi to inject the API key into the user's browser so it gets picked up by Swagger.
I don't actually know where this comes from or if there is a better solution. I just know that it works.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/urls.py
oudeismetis: hmmm yea good point. I've been bouncing back and forth on my thoughts for this. Often API docs are public by default (and I plan with a follow-on PR, maybe this PR here, to lock down the user detail endpoints by default).
That said, I could see many/most users wanting it only accessible for logged-in users.
Conflicted...

oudeismetis: I'm going to undo my change here. I think there is a version of the bootstrapper where it's focused on "I'm building an API" and settings like this should default to open. But that's currently a distant 3rd place priority.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -69,7 +68,6 @@ class UserViewSet(
     mixins.ListModelMixin,
     mixins.UpdateModelMixin,
 ):
-    swagger_schema = None
```
oudeismetis, line 72: Removing these for the same reason I mentioned elsewhere.
Yes, for a pure API app we might want to hide these so users reading the docs only get API endpoints they can actually call.
But...by default these docs are for frontend devs consuming the backend, so they shouldn't be hidden.
This has the side effect that for some cases you have to do extra work to get Swagger looking the way you want.

whusterj (approved):
LGTM 💸

PR #134, opened 2022-11-29
React - Vite, Tailwinds instead of Chakra, Graphql now optional

### LINKED ISSUE #133: React Client improvements
# Proposed
- [x] Upgrade to react 18
- [x] Move off of CRA in favor of Vite
- [x] Update Typescript version
- [x] Add basic .eslintrc and .prettierrc so that CI can run on this client

# Moving to react 18
Unless we're using packages tied to react <18 (we can also upgrade them along with react!) we could definitely upgrade to the 18th version, it's been around for a while and seems stable enough.

# Vite instead of CRA
CRA has been great for creating apps when there was no other big alternatives around, but has fallen behind both in performance and configurability compared to newer tools such as Vite.

The challenge in this step is probably set up a couple of things (such as jest, eslint) that  CRA brings out of the box while Vite does not
# Upgrade TS version
Typescript has just released 4.9 whereas we're using 3.9 in the project, we're a full major behind. 

Hopefully this upgrade is seamless since there isn't that much code

### LINKED ISSUE #135: Drop Chakra UI from React client
I'm wondering whether having chakra-ui as default in React client is a good fit.

Feels like Chakra ui has a lot of functionalities (and overhead to learn specifics of the library itself) that many new projects might not need from the very beginning.

I see the Vue client does not have any component library as default but instead it has tailwind, which I believe is a perfect styling starter for any project. 

If a given project strives for a component library then that specific project could add chakra-ui for their use.

PS: I've also had some trouble with testing components that used Chakra ( #133 has some links in the comments) so I might be a little bit triggered and biased regarding chakra, I'd love your thoughts here!

### LINKED ISSUE #139: React client: Add auth guard rather than using a public routes array
Rather than creating a separate array for specifying which are the public routes. We can leave public routes to be at the top of the router declaration, and add a Private routes auth guard instead. Similar to what we did in the The Drop:

Declaring private routes:

https://github.com/thinknimble/share-the-drop-web/blob/0d8be46067bfa169923f4f7506b1f6a1dc428eae/src/routes/index.tsx#L32-L89

Declaring Root Routes (thus, public)
https://github.com/thinknimble/share-the-drop-web/blob/0d8be46067bfa169923f4f7506b1f6a1dc428eae/src/routes/index.tsx#L91-L106

### LINKED ISSUE #143: React Client - add axios alternative to apollo when using Rest
When the project is not using gql, it makes more sense to include axios instead of apollo. 

Since we're not differentiating with using gql/rest on React client creation, it would be nice to have the placeholders for both approaches.

lakardion:
- Remove CRA and migrate project to Vite
- Remove jest and use vitest
- upgrade typescript to latest stable version
- upgrade React to v18
- Drop chakra and add tailwind
- Add better separation of private routes and public routes in a more native way with react-router 
- Dissociate gql and rest alternatives for react by adding cookiecutter templating to react code.

## What this does

Closes #133 
Closes #135 
Closes #139 
Closes #143 

## TODO
- [x] Configure jest and react testing library

~Configured `vitest` but tests are still failing due to issue with `chakra-ui`, details on the closing issue ^
(btw they were failing even before this PR)~ Chakra was removed

~@lakardion I added DO NOT MERGE to this, we first will need to make a separate PR into yours to make sure that we also change the appropriate configurations in the `settings.py` file!~

@lakardion LEGO it's time to merge this bad boy in!


lakardion:
Linting is failing due to `auth.tsx`. These issues are going to be fixed by #107

cookiecutter.json
```diff
@@ -10,10 +10,22 @@
   "_copy_without_render": [
     "clients/vue3/src/views/dashboard",
     "clients/vue3/src/components",
-    "clients/react/src/pages",
+    "clients/react/src/components",
+    "clients/react/src/pages/app-or-auth.tsx",
```
lakardion, line 14: Does someone know a way to do "all files in `react/src/pages` except `log-in.tsx` and `sign-up.tsx` ?

I tried doing
```
//...
"clients/react/src/pages",
"!clients/react/src/pages/log-in.tsx",
"!clients/react/src/pages/sign-up.tsx"
//...
```

But that did not work

paribaker: @oudeismetis

lakardion: Ah I see your point, yeah the problem with JS/TS is that if for some reason you have anything that is surrounded with `{{  }}` it throws some templating error when executing cookiecutter. 

So I preferred to leave out the pages that don't have any cookiecutter templating syntax.

It was fine before, since none of the files within `/pages`  had templating syntax there, but now I need those two to be included

paribaker: ahh yes that's true!

paribaker: It seems there was only one file that needed it so I just escaped the variable with quotes and it worked. I also went ahead and updated the package.json and settings.py for the deployment.

oudeismetis: sounds like you figured it out.
For future reference I believe the cookiecutter docs have details on how to handle complicated situations like this.

oudeismetis:
Is there a more specific goal to this PR other than "React improvements"?
I worry this might be getting scope creeped and thus harder to review and merge

{{cookiecutter.project_slug}}/clients/react/README.md
```diff
@@ -1,58 +1,46 @@
...
-feature. However we understand that this tool wouldn’t be useful if you couldn’t
-customize it when you are ready for it.
-
-## Learn More
-
-You can learn more in the
-[Create React App documentation](https://facebook.github.io/create-react-app/docs/getting-started).
-
-To learn React, check out the [React documentation](https://reactjs.org/).
+Tn-spa-bootstrapper react client
```
oudeismetis, line 3: once the project is generated it's no longer a "tn-spa-bootstrapper" project in my mind. So I'd probably leave this as just `React Client`

{{cookiecutter.project_slug}}/clients/react/README.md
```diff
@@ -1,58 +1,46 @@
...
+
+```diff
+{
+  "scripts": {
+-    {% if cookiecutter.client_app == 'React' %}
+    "build": "cd client && npm install && npm run build"
+-    {% else %}
+    "postinstall": "cd client && npm install",
+    "heroku-postbuild": "cd client && npm run heroku-postbuild"
+-    {% endif %}
```
oudeismetis, line 19: odd. Once the project is generated all this shouldn't exist in the code

lakardion: True, I got messed up between bootstrapper documentation and resulting app readme.
Anyway this does not apply anymore now that there are different paths for gql and rest. 
I used to run the project from this codebase, but now to run it we need to generate the app with cookiecutter.

I'll revisit this README

{{cookiecutter.project_slug}}/clients/react/README.md
lakardion: updated

lakardion:
> Is there a more specific goal to this PR other than "React improvements"? I worry this might be getting scope creeped and thus harder to review and merge @oudeismetis 

I guess the main trigger for the PR was the migration from CRA to Vite.

But then some other things were addressed as I received some requests:
- Add tn-forms (with the package we created for react)
- Minor rework on routes to have better definition on public vs private routes.
- Add Rest | GQL code paths
- Removal of Chakra UI in favor of tailwindcss

cookiecutter.json
paribaker: Yea but they do not offer a way to set negative look ups which sucks

paribaker:
@oudeismetis 
You are right we actually changed quite a bit, Damian and I have bootsrapped 3-4 apps recently and have set them all up in this way. We switched out to Vite and made the default app a DRF app. If someone wants to add GQL this is still possible with these updates. 

On another note, when I click on view deployment it seems to show the Vue app not the react, is that expected?

oudeismetis:
In the future we should make PRs like this 3 separate PRs as this one get a little bloated (and thus dragged on when we could have merged sooner)

I see a couple of lingering issues here with flake8, etc.
Tomorrow if those small items aren't addressed I'll pull this PR down and help out with fixing those things.

Sounds like this PR is well tested so I think we can/should get this merged before the end of the week.

{{cookiecutter.project_slug}}/clients/react/package.json
```diff
@@ -3,42 +3,62 @@
...
-    "extends": "react-app"
+    "start": "vite",
+    "build": "tsc && vite build --base=/static/",
+    "cypress": "cypress open --env TEST_USER_EMAIL=example@cypress.com,TEST_USER_PASS='pfi3n6iw##4%i3coruuf4*khg05a10t9j=l!&)6x37@=wj9n*o'",
+    "test": "vitest run src",
+    "test:watch": "vitest src",
+    "test:single": "vitest $0",
+    "prettier": "prettier --write ./src",
+    "lint": "eslint --ext .js,.ts,.jsx,.tsx src",
+    "dev-rest": "mv src/main.rest.tsx main.tsx",
```
paribaker, line 58: @lakardion  these scripts arent needed anymore right?

lakardion: good call. I forgot about these. I'll remove them

{{cookiecutter.project_slug}}/clients/react/src/pages/sign-up.tsx
```diff
@@ -0,0 +1,194 @@
...
+      logIn({
+        variables: {
+          email: data.createUser.user.email,
+          password: form.confirmPassword.value,
+        },
+      })
+    },
+    onError: (error: { message: string }) => {
+      if (error.message.includes('value too long')) {
+        //TODO: what is this?? there's no phone in this form
```
oudeismetis, line 58: If it's not used, I would delete it.

{{cookiecutter.project_slug}}/clients/react/package.json
```diff
@@ -3,42 +3,58 @@
...
-    "cypress": "source ../.env && npx cypress open",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject",
-    "heroku-postbuild": "npm build"
-  },
-  "eslintConfig": {
-    "extends": "react-app"
+    "start": "vite",
+    "build": "tsc && vite build --base=/static/",
+    "cypress": "cypress open --env TEST_USER_EMAIL=example@cypress.com,TEST_USER_PASS='pfi3n6iw##4%i3coruuf4*khg05a10t9j=l!&)6x37@=wj9n*o'",
```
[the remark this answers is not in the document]
lakardion: Whoops I think @paribaker removed the test for login though so this probably should not be here anymore.
I had added this here because cypress was not picking up the .env file for some reason when I tried that locally

{{cookiecutter.project_slug}}/clients/react/src/utils/get-cookie.js
```diff
@@ -7,17 +7,17 @@
...
+    while (c.charAt(0) === ' ') {
+      c = c.substring(1)
     }
     if (c.indexOf(name) === 0) {
-      return c.substring(name.length, c.length);
+      return c.substring(name.length, c.length)
     }
   }
-  return "";
+  return ''
```
oudeismetis, line 22: Unless there is a particularly strong reason to drop double questions and trailing semi-colons, I'd say we should NOT be doing that in this PR. I've seen several of our projects ping-pong back and forth between the two. We should pick one and stick with it.

{{cookiecutter.project_slug}}/clients/react/tests/e2e/specs/test-login.cy.ts
```diff
@@ -0,0 +1,6 @@
+describe('Tests login workflow', () => {
+  it('Home page auto redirects to login', () => {
+    cy.visit('/')
+    cy.url().should('include', '/log-in')
+  })
+})
```
oudeismetis, line 6: This is not a real test of login. Looks like you removed the already existing test that does test login.

{{cookiecutter.project_slug}}/clients/react/src/utils/get-cookie.js
lakardion: We're adding prettier and eslint in here as well.
I used the same prettier configuration we're using in the vue app:
https://github.com/thinknimble/tn-spa-bootstrapper/blob/edc1627c54162f05fb21d2f6d409893a9919df15/%7B%7Bcookiecutter.project_slug%7D%7D/clients/vue3/package.json#L70-L75

{{cookiecutter.project_slug}}/clients/react/vite.config.ts
```diff
@@ -0,0 +1,32 @@
...
+      {% if cookiecutter.use_graphql == 'n' -%}
+      proxy: {
+        '/api/': {
+          target: (env.VITE_DEV_BACKEND_URL || 'http://server:8000') + '/api',
+          changeOrigin: true,
+          rewrite: path => path.replace(/^\/api\//, ''),
+        },
+      },
+      {% endif -%}
+      port: 8089
```
lakardion, line 29: @paribaker

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -149,5 +149,9 @@ SMTP_VALID_TESTING_DOMAINS='thinknimble.com'
 
 # Testing (NOTE: Heroku and Github Actions will need to have matching values for some of these)
 DJANGO_SUPERUSER_PASSWORD='!!!DJANGO_SECRET_KEY!!!'
-CYPRESS_baseUrl='http://localhost:8080'
 CYPRESS_TEST_USER_PASS='!!!DJANGO_SECRET_KEY!!!'
+{%- if cookiecutter.client_app == "React" %}
+CYPRESS_baseUrl='http://localhost:5173'
```
lakardion, line 154: We were using 8089 for local

cookiecutter/react_template.yaml
```diff
@@ -1,3 +1,3 @@
 default_context:
```
paribaker, line 1: @oudeismetis I changed this manually for now since nobody has touched the gql in a while AND we do not have a way of differentiating I think this is ok?

oudeismetis: so...this PR broke graphql?

oudeismetis:
code style changes got merged in a separate PR. I can help fix this PR since now it's a big mess because of that.

paribaker:
I was worried about that, but I think we just need to accept the current changes since the config is the exact same right?

lakardion:
Fixing conflicts...

Yes they're all mostly old files

cookiecutter/react_template.yaml
paribaker: It's likely it did,but I would like to fix that separately.

paribaker: We have an upcoming project to start in March I'd like to have this ready by then if possible. 
Don't think I've seen a gql project started since Basta.

{{cookiecutter.project_slug}}/.env.example
paribaker: Okay Ty! That change must have occurred after Julep

paribaker:
@oudeismetis flake8/black/isort linting errors have been fixed, automated tests are working. The only issue right now is the gql, seems the url is not rendering correctly for the api. 

I think we should fix that elsewhere. Thoughts?

oudeismetis:
I think this is now ready to merge(?) Approving but leaving it here for one of you to do the honors if you agree and don't see a lingering issue.

oudeismetis (approved):

cookiecutter/react_template.yaml
oudeismetis: OK I saw your other PR with a fix for this issue. Up to you if that fix needs to be included here or merged separately.

PR #136, opened 2022-11-30
React Client - Tn forms utils

lakardion:
## What this does
- Install `@thinknimble/tn-forms` and use it on current forms
- Install `@thinknimble/tn-forms-react`
- Use those utils on Login and Signup forms


paribaker:
@lakardion I think you just need to update the packages for me to merge this! tn-forms 3.1.1 and tn-forms-react 1.0.3

lakardion:
> @lakardion I think you just need to update the packages for me to merge this! tn-forms 3.1.1 and tn-forms-react 1.0.3
I can update the package's reference in the package.json for them to be exact version
They're already being resolved to latest since it is specified as `^3.0.0` and `^1.0.0`

PR #137, opened 2022-11-30
Add a JSON Web Token Authentication class for DRF

whusterj:
## What this does

Our current GraphQL setup uses JWTs for authentication.

This PR adds a `JSONWebTokenAuthentication` class that can be used with Django Rest Framework API Views. So the same JWT that is used to authenticate GraphQL endpoints can also be used for DRF.

_Why not use GraphQL for everything?_ - In some one-off cases, a DRF API View is faster and easier to write and does not require or fit into the types graph. In my case, I had a single purpose, data-heavy endpoint that suffered poor performance when processed by GraphQL's type system. I threw together a DRF endpoint that grabs raw data from the DB -> transforms it to a list of dicts -> JSON HTTP Response.

## How to test

I have tested this in a project I'm working on. I could write a sample view and unit test, for instance.

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -1,8 +1,7 @@
+import secrets
 from os import remove, rename
 from os.path import exists, join
-import secrets
 from shutil import copy2, move, rmtree
-
 
 END = "\x1b[0m"
 QUESTION = "\x1b[0;36m [QUESTION]: "
@@ -91,6 +90,7 @@ def remove_graphql_files():
         join("server/{{ cookiecutter.project_slug }}/core", "schema.py"),
         join("server/{{ cookiecutter.project_slug }}/core", "types.py"),
         join("server/{{ cookiecutter.project_slug }}/core", "mutations.py"),
+        join("server/{{ cookiecutter.project_slug }}/core", "jwt_auth.py"),
     ]
     for file_name in file_names:
         if exists(file_name):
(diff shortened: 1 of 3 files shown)


oudeismetis:
We do have a couple of REST endpoints that are used on the React side. I wonder if it makes sense to use this in practice now.
One worry I have with code that isn't actually used anywhere is that no one will reach for it after the project exists.

(Not sure why those tests are failing. I'll probably take a look tomorrow to make sure it isn't an instability)

whusterj:
Yeah, I could try to think of a useful endpoint that demonstrates how this works. It's more of a library function, but doesn't make sense to package. I agree discoverability will be an issue when it's actually needed. Maybe it would be better for tn-gists?

oudeismetis:
maybe tn-gists would be better.
I think code changes to this repo are at their best when they are serving sensible defaults that other devs can build on top of instead of something they have to go out of their way to use or turn on.

whusterj:
Oh, actually, "smart default" is exactly what we should do!

When using GraphQL, we should swap `TokenAuthentication` with `JSONWebTokenAuthentication` in the `REST_FRAMEWORK` config of settings.py.

I can make that change and troubleshoot the build issues.

paribaker:
@

> Yeah, I could try to think of a useful endpoint that demonstrates how this works. It's more of a library function, but doesn't make sense to package. I agree discoverability will be an issue when it's actually needed. Maybe it would be better for tn-gists?

https://github.com/thinknimble/gist

oudeismetis (approved):

PR #138, opened 2022-12-02
updated aws-configs

paribaker:
## What this does

Fixes issues with our current broken implementation of S3 and static folders. 
This PR: 
- Removes lines 306 & line 308, right now these lines break the ui of the app when USE_AWS_STORAGE is set to true. Line 306 `STATICFILES_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.StaticRootS3Boto3Storage"` sets s3 as the static directory (referencing a function in a file that does not exist) and 308 sets the url for each of these files to point to s3 `STATIC_URL = f"https://{aws_s3_domain}/static/"`. However line 314 `STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"` overrides the the variable back to whitenoise. This means the static files referenced do not exist. 
- Removes the core/handlers file and creates the utils/storages file instead. Line 305 sets the `PRIVATE_MEDIA_LOCATION` to s3 but is pointing to a function in a file that does not exist (prior to this PR)
- Line 294 is updated to check if the app is a review app in Heroku and ignores AWS storage. There are a few reasons for this. If we were to use aws storage as is we would need to create a pipeline variable for the `AWS_LOCATION` variable to save to, additionally when the tear down occurs we would need to delete the folder. So until this work is done I disabled `USE_AWS_STORAGE` for review apps.


oudeismetis:
Need rebasing and fixing tests

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -13,6 +13,7 @@
 IN_DEV = ENVIRONMENT == "development"
 IN_STAGING = ENVIRONMENT == "staging"
 IN_PROD = ENVIRONMENT == "production"
+IS_REVIEW_APP = config("HEROKU_PR_NUMBER", default=0) # 0 here will result in false PR numbers start 1+
```
oudeismetis, line 16: You had this as a bool before. Reason for changing?

paribaker: the cast was failing !

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -305,9 +306,9 @@
     # Default file storage is private
     PRIVATE_MEDIAFILES_LOCATION = AWS_LOCATION + "/media"
     DEFAULT_FILE_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.PrivateMediaStorage"
-    STATICFILES_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.StaticRootS3Boto3Storage"
+    # STATICFILES_STORAGE = "{{ cookiecutter.project_slug }}.utils.storages.StaticRootS3Boto3Storage"
     COLLECTFAST_STRATEGY = "collectfast.strategies.boto3.Boto3Strategy"
-    STATIC_URL = f"https://{aws_s3_domain}/static/"
+    # STATIC_URL = f"https://{aws_s3_domain}/static/"
     MEDIA_URL = f"https://{aws_s3_domain}/media/"
```
oudeismetis, line 312: This PR is marked Ready For Review. Reason for this code being commented out? If unneeded delete(?) Or was this a temporary change while you were testing?

paribaker: Well I left it in there because should we want to return and re-add static files to work through a cdn this line would be needed again, my only purpose in this PR was to fix the bug that was in the code and not letting us use AWS storage. Can remove if you think it is necessary

oudeismetis: Not a fan of commented out code as it has a habit of never changing. But I think it's fine for now.

oudeismetis (approved):

PR #145, opened 2023-01-16
Improvement/update tests

paribaker:
## What this does

- updates the tests to use factory boy factories 
- sets postgress as the db 


{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
```diff
@@ -1,10 +1,21 @@
-name: Unit Tests 
+name: Unit Tests
 on: [push]
 jobs:
   pytest:
     runs-on: ubuntu-latest
     env:
```
paribaker, line 6: @oudeismetis can you help me with this? The config matches all the projects I swiched to psql but the error here seems to indicate maybe the db is not created? Is there maybe another place this should go ?

{{cookiecutter.project_slug}}/.env.example
```diff
@@ -83,6 +83,11 @@ DB_USER='{{cookiecutter.project_slug}}'
 DB_PASS='!!!POSTGRES_PASSWORD!!!'
 DB_HOST='127.0.0.1'
 
+# TEST DB SETTINGS THESE MUST MATCH GH ACTIONS
```
oudeismetis, line 86: REALLY good code comments in this PR ✨ 
Not redundant or obvious and drawing the readers eyes to import nuances

oudeismetis (approved):

{{cookiecutter.project_slug}}/.github/workflows/pytest.yml
oudeismetis: for future readers...
This issue here was that the bootstrapper never runs the Github Actions for the generated project. It has it's own action for running Pytest. So a change like this has to happen in 2 places.

PR #146, opened 2023-01-24
isort ignore migration files

paribaker:
## What this does

Adds skip migrations to toml file, for me migrations are getting sorted so I figured in case this happens to anyone it would be easier to add it to the project!

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -10,4 +10,4 @@ include_trailing_comma = true
 use_parentheses = true
 force_grid_wrap = 0
 multi_line_output = 3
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -10,3 +10,4 @@ include_trailing_comma = true
 use_parentheses = true
 force_grid_wrap = 0
 multi_line_output = 3
+skip = 'migrations'
```
oudeismetis, line 13: https://pycqa.github.io/isort/docs/configuration/options.html#skip
You might want `extend_skip` instead?

oudeismetis (approved):

PR #148, opened 2023-01-27
Version Bump tn-forms and tn-models

paribaker:
## What this does

Updates tn-models and tn-forms to the latest version (technically tn-forms was fine but tn-models did not have the ^)
Either way upgraded both and set the caret to always install latest


oudeismetis (approved):
Updated the PR title for you

paribaker:
> Updated the PR title for you

LOL must have got caught in two minds!

PR #150, opened 2023-01-31
Simpler Cypress Setup

oudeismetis:
## What this does

The current Cypress setup is too complicated. This change results in the same behavior with much less code and configuration.

## Checklist
- [x] Revert to the old simple way of running
- [x] Research ways to have Github actions pause until it detects a new deploy has occurred
- [x] Implement one of the approaches researched on another project
- [x] Confirm that Cypress isn't triggered on Build success, but on Release success
- [x] Test what happens when 2 commits are pushed back-to-back
- [ ] After merge, remove `HEROKU_API_KEY` and `GITHUB_TOKEN` from Github and Heroku respectively
- [ ] After merge, turn the Cypress merge requirement back on.

## Deploy Notes

n/a

## How to test

You will either need to bootstrap a new project or test this on an existing project (PR author has done the later)

Push a commit and observe that the required Chrome and Firefox steps are yellow.
After the release phase in Heroku, observe that Chrome and Firefox are triggered to run.
Push 2 more commits (spaced ~1min apart)
Observe that when the first release phase finishes, Chrome and Firefox are only triggered on the older of the two commits and that the PR is still blocked for merging.
Observe that when the 2nd release phase finishes, Cypress is finally triggered on the newest commit.

## Edge cases

Found only one minor edge case. If you push two commits back-to-back quickly, it's possible that when Cypress runs on the first commit that it will be testing the newer version of code. I think this is a very unlikely scenario to occur and, if it does occur, is unlikely to be noticed or matter to the dev team.


.github/workflows/cypress.yml
```diff
@@ -1,42 +1,18 @@
...
-on:
-  workflow_call:
-    secrets:
-      CYPRESS_TEST_USER_PASS:
-        required: true
+on: [deployment_status]
+
 jobs:
   Setup:
+    if: github.event.deployment_status.state == 'success'
```
oudeismetis, line 6: The 2 lines added here is all you need to have Cypress run after Heroku finishes deploying ✨

.github/workflows/cypress.yml
```diff
@@ -1,42 +1,18 @@
...
-      - name: Checkout
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v2
       - run: pipx install pipenv
       - uses: actions/setup-python@v2
         with:
           python-version: '3.10'
-          cache: 'pipenv'
+          # Can't use cache because of https://github.com/actions/cache/issues/319
+          # cache: 'pipenv'
```
oudeismetis, line 15: caching doesn't work for `deployment_status` events sadly. (known bug)

.github/workflows/cypress.yml
```diff
@@ -64,7 +37,7 @@ jobs:
           NPM_CONFIG_PRODUCTION: false
         working-directory: ./my_project/client
         run: npm install
-      - name: Run against ${{ needs.Setup.outputs.BASE_URL }}
+      - name: Run against ${{ github.event.deployment_status.environment_url }}
```
oudeismetis, line 40: Removed the `jwalton/gh-find-current-pr@v1` action dependency as we no longer need to "figure out" what the URL is for the deployment. Heroku gives that to as when it creates the status event.

oudeismetis: This is why the `Setup` step above is now a LOT simpler

.github/workflows/heroku_deploy.yml
```diff
@@ -13,46 +14,17 @@ jobs:
...
           else
             echo "Switching Heroku to build with $config_file instead"
             heroku plugins:install heroku-builds
-            heroku builds:cache:purge --app=${{ needs.Setup.outputs.APP_NAME }} --confirm=${{ needs.Setup.outputs.APP_NAME }}
-            heroku config:set BUILDPACK_RUN="pipenv run cookiecutter . --config-file $config_file --no-input -f" --app=${{ needs.Setup.outputs.APP_NAME }}
+            heroku builds:cache:purge --app=${{ env.APP_NAME }} --confirm=${{ env.APP_NAME }}
+            heroku config:set BUILDPACK_RUN="pipenv run cookiecutter . --config-file $config_file --no-input -f" --app=${{ env.APP_NAME }}
           fi
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
```
oudeismetis, line 30: This Github Action used to do 2 things:
1 - Tell Heroku if it needs to build a Vue app or a React app
2 - Trigger Cypress and cause it to fail early so...(blah blah complicated logic that doesn't exist anymore)

Now all this action does is detect if the dev has touched more React files or more Vue files, and then tell Heroku to change it's build type if a change is needed.

I still don't love that this runs on EVERY commit. I may try to do something smarter, but thats a future PR.

{{cookiecutter.project_slug}}/.github/workflows/cypress.yml
```diff
@@ -66,4 +34,4 @@ jobs:
           NPM_CONFIG_PRODUCTION: false
           CYPRESS_TEST_USER_EMAIL: "cypress@example.com"
           CYPRESS_TEST_USER_PASS: {{ "${{ secrets.CYPRESS_TEST_USER_PASS }}" }}
-          CYPRESS_baseUrl: {{ "${{ needs.Setup.outputs.BASE_URL }}" }}
+          CYPRESS_baseUrl: {{ "${{ github.event.deployment_status.environment_url }}" }}
```
oudeismetis, line 37: All the comments I made above for the bootstrappers `cypress.yml` are relevant here as well

oudeismetis: buuut....since this file is even simpler now, we don't even need a `Setup` step anymore

.github/workflows/cypress.yml
paribaker: Is this a big deal at the moment ? No right our apps arent that big that it would mean a huge slow down?

{{cookiecutter.project_slug}}/Procfile
```diff
@@ -14,6 +14,6 @@ web: gunicorn {{ cookiecutter.project_slug }}.wsgi --chdir=server --log-file -
 #   3. Up-to-date build numbers shown in app (ideally auto-generated at build-and-deploy time)
 #
 # Comment the line below to disable migrations automatically after a Heroku push.
-release: python server/manage.py makemigrations --noinput && python server/manage.py migrate --noinput && ./scripts/release.sh
+release: python server/manage.py makemigrations --noinput && python server/manage.py migrate --noinput
```
paribaker, line 19: What about test data? should that not get triggered?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -420,7 +420,7 @@
 # Popular testing framework that allows logging to stdout while running unit tests
 TEST_RUNNER = "django_nose.NoseTestSuiteRunner"
 
-CORS_ALLOWED_ORIGINS = ["https://{{ cookiecutter.project_slug }}-staging.herokuapp.com", "https://{{ cookiecutter.project_slug }}.herokuapp.com"]
+CORS_ALLOWED_ORIGINS = ["https://{{ cookiecutter.project_slug|replace('_', '-') }}-staging.herokuapp.com", "https://{{ cookiecutter.project_slug|replace('_', '-') }}.herokuapp.com"]
```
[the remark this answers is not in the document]
paribaker: Shouldnt be an issue when we remove heroku default in the BS

paribaker (approved):

.github/workflows/cypress.yml
oudeismetis: It's not really a big deal but it will add time to the test running. So if you push one final commit to the bootstrapper you'll need to wait an extra minute or so.

This does NOT impact the generated projects so definitely not a huge deal in my mind.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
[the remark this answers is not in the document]
oudeismetis: Correct, hopefully we can find a way to have separate heroku scripts set values like these for the project based on what was actually created in heroku. Obviously this code is effectively guessing what the URL will be (in a deterministic, but fragile way)

{{cookiecutter.project_slug}}/Procfile
oudeismetis: Test data gets run once during review app creation. That happens in a different place than this

.github/workflows/cypress.yml
paribaker: ahhh okay good call out!

{{cookiecutter.project_slug}}/Procfile
oudeismetis: https://github.com/thinknimble/tn-spa-bootstrapper/blob/d2be249bccb8cd48584c56b29721dacb2417967f/%7B%7Bcookiecutter.project_slug%7D%7D/scripts/db_setup.sh#L6

`db_setup` is what triggers test data to be created.

oudeismetis: which is called from the `postdeploy` Heroku step.
https://github.com/thinknimble/tn-spa-bootstrapper/blob/4197058db8a941235135b58f051100b707edd238/%7B%7Bcookiecutter.project_slug%7D%7D/app.json#L47
That step only ever gets triggered during review app creation and thus never runs on staging or prod.
Which is why for brand new projects you have to run some of these commands manually once against staging.

PR #151, opened 2023-02-02
updated project toml

paribaker:
## What this does

Simplified toml file, it makes isort target the black configuration 
https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#profile


whusterj (approved):
Beautiful

PR #152, opened 2023-02-02
Set generated secrets automatically

paribaker:
## What this does

This PR adds the django admin user name and password we automatically generate to the pull_reques_template.md I think most people end up adding that in every PR and its annoying sometimes. 
It also adds them to the app.json file so they are set by default on the review and staging apps. Since none of these get used or added in prod we should be sure they aren't exposed on prod

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -81,8 +81,8 @@ def set_flag(file_path, flag, value=None):
         f.truncate()
 
 
-def set_django_secret_key(file_path):
-    return set_flag(file_path, "!!!DJANGO_SECRET_KEY!!!")
+def set_django_secret_key(file_path, value=None):
+    return set_flag(file_path, "!!!DJANGO_SECRET_KEY!!!", value=value)
 
 
 def remove_graphql_files():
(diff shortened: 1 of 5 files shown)


hooks/post_gen_project.py
```diff
@@ -81,8 +81,8 @@ def set_flag(file_path, flag, value=None):
         f.truncate()
 
 
-def set_django_secret_key(file_path):
-    return set_flag(file_path, "!!!DJANGO_SECRET_KEY!!!")
+def set_django_secret_key(file_path, value=None):
+    return set_flag(file_path, "!!!DJANGO_SECRET_KEY!!!", value=value)
```
whusterj, line 85: Instead of `value=None` as a kwarg, maybe it would be "safer" to make `value` a positional argument, because I don't think we ever want `value` to be `None`. That would help prevent misuse of this function.

whusterj (changes requested):
Small recommended change to `set_django_secret_key` - otherwise looks good to me!

hooks/post_gen_project.py
paribaker: I did that here so that we do not have to change code that is already using this function, but you are right that would be a mistake if the values did not match up in all 3 areas!

whusterj (approved):
:rubber-stamp:

oudeismetis:
Updating PR title to be more readable

PR #153, opened 2023-02-07
"review" as an environment type; re-added SECRET_KEY to app.json

paribaker:
## What this does

- Adds "review" as en environment for review apps
- Enables Collect static by default 
- 

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -14,6 +14,7 @@ IN_DEV = ENVIRONMENT == "development"
 IN_STAGING = ENVIRONMENT == "staging"
 IN_PROD = ENVIRONMENT == "production"
 IS_REVIEW_APP = config("HEROKU_PR_NUMBER", default=0)  # 0 here will result in false PR numbers start 1+
+IN_REVIEW = ENVIRONMENT == "review" or IS_REVIEW_APP
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = config("SECRET_KEY")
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -14,6 +14,7 @@
 IN_STAGING = ENVIRONMENT == "staging"
 IN_PROD = ENVIRONMENT == "production"
 IS_REVIEW_APP = config("HEROKU_PR_NUMBER", default=0)  # 0 here will result in false PR numbers start 1+
+IN_REVIEW = ENVIRONMENT == "review" or IS_REVIEW_APP
```
oudeismetis, line 17: Not sure how relevant this is as, in theory, review apps, staging, and prod should all behave identically. The only difference because different secrets in the env vars.

That said, I could see doing something like disabling or mocking a payments API for review apps. But then again, ideally staging wouldn't do that (use test account creds instead) so that a test in staging is a real test of the system.

thoughts?

I guess I'm saying....I could see special settings for review apps (this PR), but perhaps our staging apps should be running `IN_PROD`.

paribaker: Well the reason for this is also the rollbar/papertrail reporting, we already make this distinction when it comes to s3 buckets (because we do not have a way yet of clearing the s3 bucket after it is torn down) so I do not see a reason for it not to be different. 
I think sometimes we need to balance `in theory` with `in reality` and work our way to move `in reality` close to the theory. 
I like keeping some differences between prod and staging, namely - the swagger docs should not be exposed in prod but can be so in staging -- where our UI devs sometimes work against (although if they open a PR immediately then that PR would have its swagger, DEBUG=True setting which is very helpful in seeing errors that we do not see in local dev because we do indeed have differences.

oudeismetis: good point for things like S3 where it's not being added on by Heroku.

oudeismetis (approved):

PR #154, opened 2023-02-09
Fix for the favicon url in the index.html file

paribaker:
## What this does

Favicon is incorrectly referenced in index.hml


paribaker:
This might not be needed !

PR #159, opened 2023-03-06
Favicon instructions use wrong URL

oudeismetis:
## What this does

Docs for updating the favicon point to a website that doesn't create favicons the way we are expecting them. Switched the docs to point to the real one we had been using.


whusterj (approved):
🥇

PR #160, opened 2023-03-08
Clients - Add react native client

lakardion:
## What this does
- Add react native as a client extra option for bootstrapper users.
## Checklist
- [x] Add react native client template from another project
- [x] Add cookie cutter command to ask whether the user wants a react-native client or not
- [x] Add gh action files for expo. Remove them if the user chooses not to include react native project


cookiecutter.json
```diff
@@ -27,5 +27,6 @@
     "Vue3",
     "React",
     "None"
-  ]
+  ],
+  "include_mobile": "n"
```
oudeismetis, line 31: I think this could be a clearer question?
So all this will do will add in react native login? Does this require picking react above to work? If they want a mobile app should they even be allowed to pick web as well? or only have react as an option?

hooks/post_gen_project.py
```diff
@@ -56,20 +56,30 @@ def get_random_secret_key():
...
-    move(join("clients", client), join("client"))
-    rmtree(join("clients"))
+    move(join(web_clients_path, client), join("client"))
+    rmtree(join(web_clients_path))
     env_path = join("client", ".env.local.example")
     if exists(env_path):
         rename(env_path, join("client", ".env.local"))
 
+def remove_mobile_client_files(client):
+    rmtree(join(mobile_clients_path,client))
```
oudeismetis, line 76: I guess we don't have a linter setup to run on this file for the PR checks, but this file has a bunch of linting issues.

oudeismetis: not entirely your fault, so I think you could ignore for now and we could fix that in a lint-only seperate PR

{{cookiecutter.project_slug}}/.github/workflows/expo-emergency-prod-update.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: 🏗 Setup Expo and EAS
+        uses: expo/expo-github-action@v7
+        with:
+          expo-version: latest
+          eas-version: latest
+          token: {{ "${{ secrets.EXPO_TOKEN }}" }}
```
oudeismetis, line 28: The extra pair of curliest is both confusing and annoying, glad you were able to figure it out on your own 😄

{{cookiecutter.project_slug}}/.github/workflows/expo-pr.yml
```diff
@@ -0,0 +1,51 @@
...
+on: [pull_request]
+jobs:
+  Setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: jwalton/gh-find-current-pr@v1
+        id: findPr
+      - name: Set URL for PR
+        if: success() && steps.findPr.outputs.number
+        run: echo "BASE_URL=https://example-pr-${PR}.herokuapp.com/" >> $GITHUB_ENV
```
oudeismetis, line 11: check out the latest update to the Cypress Github Action
https://github.com/thinknimble/tn-spa-bootstrapper/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/.github/workflows/cypress.yml#L19

Heroku will create a deployment event, which you can use the trigger a workflow like this one. That also means it provides the URL needed to call it, so you no longer need to use `jwalton/gh-find-current-pr@v1`

I imagine right now you are having problems where this action fails unless/until a deployment has succeeded and is thus a bit unstable

{{cookiecutter.project_slug}}/clients/mobile/react-native/.gitignore
```diff
@@ -0,0 +1,88 @@
+node_modules/**/*
```
oudeismetis, line 1: hmmm...I would think a file like this would be the same for all projects

{{cookiecutter.project_slug}}/clients/mobile/react-native/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  printWidth: 100,
+  trailingComma: 'all',
+  singleQuote: true,
+  semi: false,
+  tabWidth: 2,
+};
```
oudeismetis, line 7: same here. I wonder how many files are common between both web and mobile work.

{{cookiecutter.project_slug}}/clients/mobile/react-native/README.md
```diff
@@ -0,0 +1,231 @@
...
+- "Testing React Native apps with zero effort" - [Medium](https://kanzitelli.medium.com/cli-rn-making-rn-app-developing-experience-as-smooth-as-possible-1022aae3a0d3), [Dev.to](https://dev.to/kanzitelli/cli-rn-making-rn-app-developing-experience-as-smooth-as-possible-4e98)
+- "Expo + React Native Navigation? Yes!" - [Medium](https://kanzitelli.medium.com/expo-react-native-navigation-yes-ebda0cbfa4b1), [Dev.to](https://dev.to/kanzitelli/expo-react-native-navigation-1pll)
+
+## Why
+
+...do we need yet another starter/boilerplate? Well, I work with React Native for more than 4 years and during the time I started having my own project structure which was a good fit for almost all of the delivered apps. Also, I have come up with some custom [useful services/methods](#useful-servicesmethods) which simplify usage of [React Navigation](https://github.com/react-navigation/react-navigation) and other libraries. Check out [Advantages](#advantages) section.
+
+## License
+
+This project is [MIT licensed](/LICENSE.md)
```
oudeismetis, line 231: this is not correct

cookiecutter.json
paribaker: Weird question 🤔  why would that not be allowed?
I like the idea of having the client as an option, for Julep for example we are likely going to need Webviews to work.

{{cookiecutter.project_slug}}/clients/mobile/react-native/.gitignore
paribaker: Not sure what this comment means can you elaborate?

paribaker:
@oudeismetis There are some things we are still working on figuring out that can be fixed over time. In general this should not yet be merged as we are still finding a solution for .env variables and want to try a few things out with julep before we do

cookiecutter.json
lakardion: We could do `include_mobile_project` ?
This is an independent question from web clients, you can choose whatever client or even none and still get a react native project if you answer this as `yes`

hooks/post_gen_project.py
lakardion: Oh maybe my vscode is not configured properly. I didn't get any warnings. I don't usually get into python files

{{cookiecutter.project_slug}}/.github/workflows/expo-emergency-prod-update.yml
lakardion: Happened to me as well on a React file where we had a prop such as `style={ { fontSize: 12 } }` and it was blowing up as well, so we had to do something similar than here

{{cookiecutter.project_slug}}/clients/mobile/react-native/.prettierrc.js
lakardion: Right this is the same as web's. 
I wonder whether we could reference an outer file for prettier so we could gather them together somewhere in an higher folder
Having it in the root of the project just makes it work automatically

I haven't tried this out but what do you think about this option?:
https://prettier.io/docs/en/configuration.html#sharing-configurations

We could reference it in all projects and it would be a single source of truth (an npm package that we create)

cookiecutter.json
oudeismetis: OK so you could pick Vue and mobile and everything should be fine (though I imagine most wouldn't do that)

hooks/post_gen_project.py
oudeismetis: I fixed this in a separate PR. You'll likely have a conflict now.

{{cookiecutter.project_slug}}/clients/mobile/react-native/README.md
lakardion: Removed

hooks/post_gen_project.py
lakardion: There were some other formatting issues in there so I got to use black and then the checks passed!

oudeismetis:
@lakardion Since this PR isn't going to be merged soon, I'm think the part of this PR that creates the `web` folder and moves Vue/React should be it's own PR. We can merge that quickly and that will help avoid conflicts will you iterate on this.

thoughts?

lakardion:
> @lakardion Since this PR isn't going to be merged soon, I'm think the part of this PR that creates the web folder and moves Vue/React should be it's own PR. We can merge that quickly and that will help avoid conflicts will you iterate on this.
> thoughts? @oudeismetis 
Great suggestion!. Created #165

paribaker:
@lakardion  lets fix the conflict and then tomorrow after we run this with our project we can merge or update!

{{cookiecutter.project_slug}}/.github/workflows/expo-pr.yml
paribaker: ughh this failed for me on Worthy need to try it again and will update!

oudeismetis: The `findPr` action often fails the first time a PR is opened. So you either have to push another commit or manually re-trigger it.
Another reason to move away from it and just trigger on a deploy event.

scripts/vue_or_react.sh
```diff
@@ -2,8 +2,8 @@
 
 # cookiecutter tn-spa-bootstrapper --config-file tn-spa-bootstrapper/cookiecutter/vue_template.yaml --no-input
 git fetch origin main
-react_count=$(git diff --name-only origin/main -- | grep "/clients/web/react" | wc -l)
-vue_count=$(git diff --name-only origin/main -- | grep "/clients/web/vue" | wc -l)
+react_count=$(git diff --name-only origin/main -- | grep "/clients/web/react/" | wc -l)
+vue_count=$(git diff --name-only origin/main -- | grep "/clients/web/vue/" | wc -l)
```
oudeismetis, line 6: adding the slash here will break things.
One solution is the make this `vue3` with the slash included.

hooks/post_gen_project.py
```diff
@@ -57,11 +57,12 @@ def get_random_secret_key():
...
 
-def remove_web_client_files(client):
-    rmtree(join(web_clients_path, client))
+web_clients_path = "clients/web"
+mobile_clients_path = "clients/mobile"
 
 
-web_clients_path = "clients/web"
+def remove_web_client_files(client):
+    rmtree(join(web_clients_path, client))
```
oudeismetis, line 65: I've noticed that the `clients` directory is now left behind (and empty) after generating a new project

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/axios-instance.ts
```diff
@@ -2,13 +2,13 @@ import axios, { AxiosError } from 'axios'
...
 
-let config = Config as any
+const appConfig = Config as any
 export const axiosInstance = axios.create({
-  baseURL: config?.apiUrl,
+  baseURL: appConfig?.backendServerUrl,
 })
 
 axiosInstance.interceptors.request.use(
-  async (config) => {
```
lakardion: I wonder why you need any here 🤔 . This callback is typed by axios already

paribaker (approved):

{{cookiecutter.project_slug}}/clients/mobile/react-native/.prettierrc.js
oudeismetis: yea, something to refactor later after the dust has settled and we are happy with all these changes.

oudeismetis:
I think the comment is buried, but there is still an issue with the vue_or_react script change.

scripts/vue_or_react.sh
lakardion: So we should rename the vue folder to vue3?
And this would end up being `[...]/web/vue3/` ?

paribaker (approved):

oudeismetis:
Failing because the app that is built is a vue app but cypress is expecting a react app.
I expect this is because of that bug fix in the vue_or_react script. Recreating the review app should fix it. Doing that now.

oudeismetis:
Fixed!

PR #161, opened 2023-03-09
Black and flake8 don't run on the root project

oudeismetis:
## What this does

Black runs but doesn't fail PRs when it finds this. Possible we have a similar problem for other jobs. 

## Checklist
- [x] Get black to fail (root project, vue, and react
- [x] Fix issues that black found
- [x] Get flake8 to fail on the root project
- [x] Fix flake8 issues
- [x] Disable some branch protections since renaming Github Actions causes a problem

## Deploy Notes

After merge...turn branch protections back on for Pytest and the new black/flake8 as well

## How to test

checks pass on this PR
Review app looks good

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
@@ -11,7 +11,7 @@ jobs:
           python-version: '3.10'
           cache: 'pipenv'
       - run: pipenv install --dev --deploy
-      - run: pipenv run black server
+      - run: pipenv run black --check server
   flake8:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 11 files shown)


{{cookiecutter.project_slug}}/.github/workflows/linting.yml
```diff
@@ -11,7 +11,7 @@ jobs:
           python-version: '3.10'
           cache: 'pipenv'
       - run: pipenv install --dev --deploy
-      - run: pipenv run black server
+      - run: pipenv run black --check server
```
oudeismetis, line 14: @OkayItsMikael Called this out before.
Black will run but it won't fail the PR. This fixes that.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -70,9 +70,9 @@
     "dj_rest_auth",
     "django_filters",
     "django_extensions",
-    {% if cookiecutter.use_graphql == 'y' -%}
+{% if cookiecutter.use_graphql == 'y' -%}
     "graphene_django",
-    {% endif -%}
+{% endif -%}
```
oudeismetis, line 75: yes...I know...slightly gross...but I couldn't figure out a way for it remove lines in a way that wouldn't result in a linting error

paribaker (approved):

PR #162, opened 2023-03-13
HTML emails with styling

oudeismetis:
## What this does

Default templates for HTML emails now updated to be nicer looking by default:
image

Leverages the well tested [Mailgun Templates](https://github.com/mailgun/transactional-email-templates) so they should work on most any email client out of the box. 

## Other things:

- Template inheritance so devs don't have to look at the ugly HTML
- premailer library to inline styles


paribaker (approved):
Awesome, do you think maybe a quick comment in the readme to explain to folks how to change the logo is in order? I know sometimes understanding how static files work in django is a pain

paribaker:
@eliasbiagioninc  can you use this diff to add these changes to Juelp emails! I think for worthy we will just do it with a full upgrade!

oudeismetis:
yea agree, I imagine most will forget to update this image until later when they are finally putting eyes on a real email

oudeismetis:
@paribaker
- README update for logo locations
- rebased off main

paribaker (approved):

PR #163, opened 2023-03-14
Fix small issues that confuse new devs

oudeismetis:
## What this does

Started a new project recently with devs who hadn't seen this tech stack before. Ran into several points of confusion. Adding those improvements here.

## Checklist
- [x] PR template
- [x] Command and docs to seed Docker with test data
- [x] Improve root README with more steps
- [x] Better detail descriptions for `.env.local.example`
- [x] Heroku/Cypress have different defaults when React/Vue change count is the same

## Deploy Notes

Add any notes for changes needed when deployed. I.e. API keys, environment variables, database changes, etc.

## How to test

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,18 +1,12 @@
 ## What this does
 
-Add a description of what feature/page this PR intends to add or change. Don't forget to start the PR title with the Jira ticket number.
-
-- [Jira ticket](Paste Jira link here) 
+Add a description of what feature/page this PR intends to add or change. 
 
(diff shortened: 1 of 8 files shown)


oudeismetis:
Something weird is going on in Heroku. This PR doesn't have any crazy changes. 

The builds are failing with a reference to react-native. I suspect the cache is shared between our PRs on the Heroku side. Normally that wouldn't be a problem but right now there is another PR open that adds react-native.

paribaker (approved):

PR #165, opened 2023-03-15
Move React/Vue to a web subfolder in prep for react-native

lakardion:
Preparing bootstrapper for incoming addition of react-native client: #160 

## What this does
- move clients to a web folder
- make required changes in cookiecutetr
- make required changes wherever directory was referenced


oudeismetis:
This is going to fail because of a cache issue on Heroku. Our other PR with the react-native work is polluted the cached node_modules on Heroku.

I have another PR open with this problem.

Let me look at the code. If it's obvious enough I think we can temporarily ignore Heroku

oudeismetis (approved):

PR #169, opened 2023-03-30
add cookie to axios headers

villagrat:
## What this does
1. Use TS on `utils/get-cookie`
2. Add cookie to `X-CSRFToken` Header

## Checklist
- [ ] Check utility is correctly adding CSRF Header on review app
- [ ] Check created cookiecutter app behaves the same
- [ ] Run all tests

## How to test
Testable in the [PR review app](https://tn-spa-bootstrapper-pr-169.herokuapp.com/)


{{cookiecutter.project_slug}}/clients/web/react/src/utils/get-cookie.ts
```diff
@@ -6,12 +6,12 @@
  * Usage:
```
lakardion, line 6: Love these changes. Especially the named export over default!

lakardion (changes requested):

{{cookiecutter.project_slug}}/clients/web/react/src/utils/get-cookie.ts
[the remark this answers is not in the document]
villagrat: Nice catch. I'm on it

lakardion: But beware :

https://github.com/thinknimble/tn-spa-bootstrapper/blob/4c5cc6d638cb4a433013293600204923c865154a/%7B%7Bcookiecutter.project_slug%7D%7D/clients/web/react/src/services/apollo-client.ts#L4

^ let's update this to be named import!

lakardion (approved):

hooks/post_gen_project.py
```diff
@@ -114,7 +114,6 @@ def remove_gql_react_files():
     file_names = [
         join("client/src/utils", "mutations.ts"),
         join("client/src/utils", "queries.ts"),
-        join("client/src/utils", "get-cookie.js"),
```
oudeismetis, line 117: Is this used on the Vue side then?

oudeismetis: ahhh saw other comment

lakardion: Nope. That fn only runs post gen of a react project. So `client/src/utils` is from react

PR #172, opened 2023-04-14
Update for SERVER_EMAIL variables

eliasbiagioninc:
## What this does

This update avoid having hardcoded and fixed emails account in the `settings.py` file for ``SERVER_EMAIL`` variable. Instead, It takes the value from the `.env` file

## Checklist
- [x] Add `DEFAULT_FROM_EMAIL` variable to `.env` file, to set custom email account
- [x] Update `SERVER_EMAIL` variable to take the value from `DEFAULT_FROM_EMAIL` env variable


oudeismetis (approved):
Keeping code separate from configuration 👍

paribaker (approved):

PR #176, opened 2023-04-26
Feature/mobile review app

paribaker:
## What this does

This PR adds a mobile deployment workflow for the bootstrapper, when a PR is opened a new mobile app (expo go) will be published as well (assuming the mobile directory is touched)
It also makes some changes to the eas.json and app.config.json to make variable substitution more predictable in a PR

## In next pr 
- instead of copying eas.json and app.json from a template replace the variables
- add missing variables to eas update for staging 
- create a merge to main pipeline as well 
- create gh actions for a manual deploy to testflight and play store


{{cookiecutter.project_slug}}/.github/workflows/expo-main.yml
```diff
@@ -4,9 +4,31 @@ on:
...
+          echo "BUILD_MOBILE_APP=$(git diff-tree --no-commit-id --name-only -r {{ "${{ github.sha }}" }} | xargs | grep mobile/ | wc -l)" >> $GITHUB_OUTPUT
 
+  printouts:
+    runs-on: ubuntu-latest
+    needs:  setup
+    steps:
+      - name: Echo files
+        run: |
+          echo "{{ "${{ needs.setup.outputs.BUILD_MOBILE_APP}}" }}"
+          echo "$([[ needs.setup.outputs.BUILD_MOBILE_APP -gt 0 ]] && echo 'Changes detected in mobile dir..' || 'No changes in mobile app detected... will not deploy to expo')"
```
paribaker, line 27: -gt here is incorrect

.github/workflows/cypress.yml
```diff
@@ -80,3 +80,77 @@ jobs:
           CYPRESS_TEST_USER_EMAIL: "cypress@example.com"
           CYPRESS_TEST_USER_PASS: ${{ secrets.CYPRESS_TEST_USER_PASS }}
           CYPRESS_baseUrl: ${{ github.event.deployment_status.environment_url }}
+
+  configuremobile:
+    needs: Setup
+    runs-on: ubuntu-latest
+    outputs:
```
paribaker, line 87: @oudeismetis can you see what the issue with this is?
It says 
`fatal: bad revision 'origin/main'`

paribaker:
@oudeismetis I put the mobile review app in the cypress.yml for now since it does the same work but will separate that out if you think it is necessary!

.github/workflows/cypress.yml
```diff
@@ -80,3 +81,81 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 5
+        name: check diff
+      - id: checkdiff
+        run: |
+          git fetch origin main
+          echo $(git diff --name-only origin/main -- | grep "/clients/mobile/react-native/" | wc -l)
+          echo "BUILD_MOBILE_APP=$(git diff --name-only origin/main -- | grep "/clients/mobile/react-native/" | wc -l)" >> $GITHUB_OUTPUT
```
oudeismetis, line 99: I think this is fine for now but we could just improve the `vue_or_react.sh` script to be more generic and just have it spit out "Whats the primary stack this PR is about". Then do cleaner `if` guards elsewhere in this file.

But again, not needed. Can refactor later.

.github/workflows/cypress.yml
```diff
@@ -80,3 +81,81 @@ jobs:
...
+        working-directory: ./my_project/mobile
+        run: |
+          eas update --branch="${{ needs.getprnumber.outputs.PR }}" --non-interactive --auto
+          echo "${{ env.BACKEND_SERVER_URL }} and ${{ needs.getprnumber.outputs.PR }}"
+        env:
+          BACKEND_SERVER_URL: "${{ github.event.deployment_status.environment_url }}"
+          ROLLBAR_ACCESS_TOKEN: "1a19e5da05b2435b802d5a81aba2bbd7"
+          SENTRY_PROJECT_NAME: "tn-staging"
+          SENTRY_DSN: "https://a7cea97f07ac42fa9e28800b037997c9@o4504899535962112.ingest.sentry.io/4504906332897280"
+          SENTRY_AUTH_TOKEN: "${{ secrets.SENTRY_AUTH_TOKEN }}"
```
oudeismetis, line 161: OK yea....so this isn't about Cypress at all. You are just sticking this here.
I don't love it but can live with it for the moment as long as we refactor it soon(ish).

.github/workflows/heroku_deploy.yml
```diff
@@ -26,5 +27,6 @@ jobs:
             heroku builds:cache:purge --app=${{ env.APP_NAME }} --confirm=${{ env.APP_NAME }}
             heroku config:set BUILDPACK_RUN="pipenv run cookiecutter . --config-file $config_file --no-input -f" --app=${{ env.APP_NAME }}
           fi
+          cat $config_file
```
oudeismetis, line 30: a new echo and cat. Both just for logging insight?

cookiecutter/vue_template.yaml
```diff
@@ -1,3 +1,4 @@
 default_context:
   client_app: "Vue3"
   use_graphql: "n"
+  include_mobile: "n"
```
oudeismetis, line 4: not needed if it defaults to `n` anyway

scripts/vue_or_react.sh
```diff
@@ -4,9 +4,18 @@
...
-  echo "cookiecutter/vue_template.yaml"
+  config_file_path="cookiecutter/vue_template.yaml"
 fi
+
+if [ "$rn_count" != 0 ]; then 
+  cp -r resources/ {{cookiecutter.project_slug}}/
+  sed -i.bak 's/include_mobile: "n"/include_mobile: "y"/' $config_file_path
+fi
+
+echo $config_file_path
```
oudeismetis, line 21: So you did improve this file then...
I don't love the use of sed here. I feel like it would be cleaner if this script just tells you what the focus of the PR is about and then with that info we pick which tempate to use from `cookiecutter/{script_result}_template.yaml`

{{cookiecutter.project_slug}}/.github/workflows/expo-pr.yml
```diff
@@ -16,6 +15,7 @@ jobs:
       PR: {{ "${{ env.PR }}" }}
 
   publish:
+    if: github.event.deployment_status.state == 'success' # guarantee backend deployment first
```
oudeismetis, line 18: why move this?

.github/workflows/heroku_deploy.yml
paribaker: Yes please it was super helpful understanding when/why the mobile app was/wasn't building

.github/workflows/cypress.yml
paribaker: maybe we can move the artifacting to a parent yaml file and then call the various ones -- cypress/mobile deploy based on the script!

.github/workflows/cypress.yml
paribaker: Yea sticking it in here because of all the pre work that was done, building the app on heroku, uploading an artifact and then downloading it

{{cookiecutter.project_slug}}/.github/workflows/expo-pr.yml
paribaker: @oudeismetis is it going to cause a problem? I moved it because the successful deployment does not need to occur to find the backend PR. 
I guess leaving it in there also avoids running that one as well

scripts/vue_or_react.sh
paribaker: @oudeismetis  I tried to do the least to affect the rest of the builds I also did not want to remove building a client altogether because I do know there is a syntax bug when no client is selected (I believe one line loses too many spaces so the whole app crashes on building)

lakardion (approved):

lakardion (approved):

PR #178, opened 2023-05-19
Small improvements: init_db.sh, README, and logging

oudeismetis:
## What this does

A bunch of small quality of life improvements (see list below)

## Checklist
- [x] The `init_db.sh` script will now delete an existing DB if it finds one
- [x] `README` now has a default table of auto-created accounts, creds, what access they have, etc.
- [x] Verbose logging turned on by default so log messages will tell you the file and line number a message came from
- [x] Added default/example log messages to a couple of places

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -293,7 +293,7 @@ if config("USE_AWS_STORAGE", cast=bool, default=False) and not IS_REVIEW_APP:
     AWS_STORAGE_BUCKET_NAME = config("AWS_STORAGE_BUCKET_NAME")
     AWS_SECRET_ACCESS_KEY = config("AWS_SECRET_ACCESS_KEY")
     AWS_S3_CUSTOM_DOMAIN = AWS_STORAGE_BUCKET_NAME + ".s3.amazonaws.com"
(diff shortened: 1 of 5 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -293,7 +293,7 @@
     AWS_STORAGE_BUCKET_NAME = config("AWS_STORAGE_BUCKET_NAME")
     AWS_SECRET_ACCESS_KEY = config("AWS_SECRET_ACCESS_KEY")
     AWS_S3_CUSTOM_DOMAIN = AWS_STORAGE_BUCKET_NAME + ".s3.amazonaws.com"
-    AWS_LOCATION = config("AWS_LOCATION", default="")
+    AWS_LOCATION = config("AWS_LOCATION", default="")  # production, staging, etc
```
paribaker, line 296: @oudeismetis actually can we let this error? We had a problem in Lighten where I did not set it and so we kept getting rollbar errors that were indicating the error was with a trailing slash but really it was because Location was missing!

paribaker (approved):

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: yea we can switch to that. This is a pretty critical setting which is why I added the code comment.

paribaker (approved):

PR #185, opened 2023-06-06
404 on bad routes, pytest warnings, and prod log level

oudeismetis:
## What this does

A couple of small improvements

## Checklist
- [x] Improving logging
- [x] Pytest config should only ignore certain warnings
- [x] Default django template should be a 404

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
@@ -23,7 +23,7 @@ jobs:
           cache: 'pipenv'
       - run: pipenv install --dev --deploy
       - run: pipenv run flake8 server
-  {% if cookiecutter.client_app == 'Vue3' or cookiecutter.client_app == 'React' %}
+{%- if cookiecutter.client_app == 'Vue3' or cookiecutter.client_app == 'React' %}
   ESLint_and_Prettier:
     runs-on: ubuntu-latest
     steps:
@@ -39,10 +39,9 @@ jobs:
           npm install
       - name: Run ESLint
         working-directory: ./client
-        {% if cookiecutter.client_app == 'Vue3' %}
+        {% if cookiecutter.client_app == 'Vue3' -%}
         run: npx eslint --ext .js,.vue src
-        {% elif cookiecutter.client_app == 'React' %}
+        {% elif cookiecutter.client_app == 'React' -%}
         run: npx eslint --ext .ts,tsx,js src
-        {% endif %}
-  {% endif %}
-
+        {% endif -%}
+{%- endif %}
--- a/{{cookiecutter.project_slug}}/pytest.ini
+++ b/{{cookiecutter.project_slug}}/pytest.ini
@@ -1,5 +1,7 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = {{ cookiecutter.project_slug }}.test_settings
 python_files = tests.py test_*.py *_tests.py
-addopts = -p no:warnings --strict-markers --no-migrations
+addopts = --strict-markers --no-migrations
 mccabe-complexity=10
+filterwarnings =
+    ignore::DeprecationWarning
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
@@ -5,6 +5,7 @@ from django.template.exceptions import TemplateDoesNotExist
 from django.template.response import TemplateResponse
 from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import ensure_csrf_cookie
+from rest_framework import status
 {% endif -%}
 
 
@@ -18,7 +19,7 @@ def index(request):
 
 def index(request):
     try:
-        return render(request, "index.html")
+        return render(request, "index.html", status=status.HTTP_404_NOT_FOUND)
     except TemplateDoesNotExist:
-        return render(request, "core/index-placeholder.html")
+        return render(request, "core/index-placeholder.html", status=status.HTTP_404_NOT_FOUND)
 {% endif -%}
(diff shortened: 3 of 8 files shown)


{{cookiecutter.project_slug}}/.github/workflows/linting.yml
```diff
@@ -39,10 +39,9 @@ jobs:
...
+        {% if cookiecutter.client_app == 'Vue3' -%}
         run: npx eslint --ext .js,.vue src
-        {% elif cookiecutter.client_app == 'React' %}
+        {% elif cookiecutter.client_app == 'React' -%}
         run: npx eslint --ext .ts,tsx,js src
-        {% endif %}
-  {% endif %}
-
+        {% endif -%}
+{%- endif %}
```
oudeismetis, line 47: Running the project for the first time in a while.
There were a number of small regressions introduced in recent PRs that added extra whitespace and broke linters

{{cookiecutter.project_slug}}/pytest.ini
```diff
@@ -1,5 +1,7 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = {{ cookiecutter.project_slug }}.test_settings
 python_files = tests.py test_*.py *_tests.py
-addopts = -p no:warnings --strict-markers --no-migrations
+addopts = --strict-markers --no-migrations
 mccabe-complexity=10
+filterwarnings =
+    ignore::DeprecationWarning
```
oudeismetis, line 7: Pytest will no longer suppress ALL warnings in it's logs.
The `DeprecationWarning` is loud and distracting, so disabling that one.
On another project I found that some important warnings that would help you find bugs were being hidden.

This should result in the pytest error logs being a bit more useful than they were before.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
```diff
@@ -18,7 +19,7 @@ def index(request):
 
 def index(request):
     try:
-        return render(request, "index.html")
+        return render(request, "index.html", status=status.HTTP_404_NOT_FOUND)
     except TemplateDoesNotExist:
-        return render(request, "core/index-placeholder.html")
+        return render(request, "core/index-placeholder.html", status=status.HTTP_404_NOT_FOUND)
```
oudeismetis, line 24: You'll still get the index placeholder returned, but that status for that return will now be a `404` instead of a `200`.
This is particularly useful when writing some tests. If your URL was bad (ex: extra `/`) and you don't have a response object to inspect, your test would pass since you got a 200 on a URL route that doesn't exist.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -357,7 +357,6 @@
     "handlers": {
         "console": {
             "level": "INFO",
-            "filters": ["require_debug_true"],
```
oudeismetis, line 360: The only possible controversial change here.
By default we were effectively turning off logs in production. Some things would get logged, but there would be missing details.
In my mind a project should start with very verbose logs in prod. If the devs find that too noisy they can turn them off later when that becomes a problem.

oudeismetis:
weird. Not sure why the tests are failing. Saying it's getting on 404 on the review app home page.
I assume it's fluke. Will rerun after rebasing with the next PR merger.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
whusterj: Ah, this must have been mangled along the way. The "console" handler in Django's default logging requires DEBUG=True, because it has a separate "mail admins" config for production when DEBUG=False. You can see here: https://github.com/django/django/blob/main/django/utils/log.py#L18

Long ago, I overloaded the default logging config here in settings.py so that we could always have the 'django.requests' and 'django.tasks' sub-loggers below and to also make some customizations to the formatters. This is definitely worth a revisit now.

Also, I still see logs in production for apps using this config (like Squirrel), so that's interesting...

oudeismetis: yea I made this change a bit ago so I kinda forgot the details. I do recall logs still being in prod, but they weren't full detail. Definitely not using the verbose formatter, so you wouldn't get things like line numbers and I think maybe INFO logs weren't showing up(?)

Interesting context. Honestly I also don't love a change like this because it's not obvious how a future dev is to know to add `require_debug_true` as a filter should they want to change the behavior. I think many devs would end up googling their own solution.

whusterj: OK, this is probably worth a GH issue someone can tackle to test and tweak this in more detail to something we like, but for now I'm on board with this change.

For now it might also be good to add a link to Django's logging docs, in case anyone is wondering what the heck is going on here.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
```diff
@@ -1,14 +1,15 @@
...
-from django.template.exceptions import TemplateDoesNotExist
-{% else -%}
+{% if cookiecutter.use_graphql == "y" -%}
 from django.template.response import TemplateResponse
 from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import ensure_csrf_cookie
+{% else -%}
+from django.shortcuts import render
+from django.template.exceptions import TemplateDoesNotExist
+from rest_framework import status
```
oudeismetis, line 8: Just moving things around to keep it clean. The top had graphql 2nd but the code below had it first. Also making both affirmative checks and consistent quoting.

mayakeeley (approved):
lgtm!! tests failing appears to be unrelated to this work

Issue #186, opened 2023-06-08
Missing env var in .env.example


paribaker:
We recently added a new env var 
`ENABLE_EMAILS`

but did not include it in the .env.example, while it defaults to `False` its best to add it as an example to prompt users

https://github.com/thinknimble/tn-spa-bootstrapper/blob/ca60c3534a1480c8edc28835c0ed7a3d06942228/%7B%7Bcookiecutter.project_slug%7D%7D/.env.example#L138

MatterAndy:
Hello. I just made PR #187 to resolve this issue. Hope this helps!

oudeismetis:
Resolved in https://github.com/thinknimble/tn-spa-bootstrapper/pull/187
Thanks @MatterAndy !!

PR #187, opened 2023-06-08
Added ENABLE_EMAILS env var to .env.example.

MatterAndy:
## What this does

Resolves issue #186 by adding the ENABLE_EMAILS env var to [.env.example](https://github.com/thinknimble/tn-spa-bootstrapper/blob/ca60c3534a1480c8edc28835c0ed7a3d06942228/%7B%7Bcookiecutter.project_slug%7D%7D/.env.example).

## Checklist
- [x] Add the ENABLE_EMAILS env var to [.env.example](https://github.com/thinknimble/tn-spa-bootstrapper/blob/ca60c3534a1480c8edc28835c0ed7a3d06942228/%7B%7Bcookiecutter.project_slug%7D%7D/.env.example)

## How to test

Check the [.env.example](https://github.com/thinknimble/tn-spa-bootstrapper/blob/ca60c3534a1480c8edc28835c0ed7a3d06942228/%7B%7Bcookiecutter.project_slug%7D%7D/.env.example) file.

### DIFF
--- a/{{cookiecutter.project_slug}}/.env.example
+++ b/{{cookiecutter.project_slug}}/.env.example
@@ -136,6 +136,7 @@ REDIS_URL='redis://127.0.0.1:6379/1'
 #
 {%- if cookiecutter.mail_service == 'Mailgun' %}
 # Mailgun
+ENABLE_EMAILS='False'
 MAILGUN_API_KEY=''
 MAILGUN_DOMAIN=''
 MAILGUN_API_URL=''


{{cookiecutter.project_slug}}/.env.example
```diff
@@ -136,6 +136,7 @@ REDIS_URL='redis://127.0.0.1:6379/1'
 #
 {%- if cookiecutter.mail_service == 'Mailgun' %}
 # Mailgun
```
paribaker, line 138: This should be before both if statements so that it can apply to both cases!

paribaker (changes requested):
The ENABLE_EMAILS variable should be put before the if statements so that it can apply to both cases (anymail or mailgun)

MatterAndy:
Sorry about that. I just moved it so that it applies to both cases.

paribaker (approved):
This looks good to merge!

Issue #167, opened 2023-03-28
Update Signals


paribaker:
Currently the signals are imported inside the models.py according to django this is not advised. 
https://docs.djangoproject.com/en/4.1/topics/signals/
One option is to import all signals in apps.py as such

```
class SomeAppConfig(AppConfig):
    default_auto_field = "django.db.models.BigAutoField"
    name = "some_project.some_app"

    def ready(self) -> None:
        # this import is required to register signals after the app is initialized
        # https://docs.djangoproject.com/en/4.1/topics/signals/
        import some_project.some_app.signals  # noqa

        return super().ready()
```
        
Alternatively a better approach is that you can import and connect to each signal @juliantn has an example

PR #191, opened 2023-06-10
move signals to signals.py and add to apps.py

paribaker:
## What this does
Moves the signal to the apps.py

https://github.com/thinknimble/tn-spa-bootstrapper/issues/167


NC-Fran (approved):
Nice improvement!

PR #192, opened 2023-06-10
Issue 188 adds alternative flow for password reset - needed for mobile (BE only)

paribaker:
## What this does

Adds two new endpoints for sending password reset codes instead of links. Currently our mobile apps do not support deep linking and so we need an alternative. 
Instead we send the user a 5 digit expiring pin code. 

**At the moment this will be included by default but we could also add this in as part of cookiecutter**

## Checklist

- [x] tests
- [x] views
- [x] models
- [x] serializers
- [x] urls
- [x] Update email to include one for code emails (I have an example)

## How to test


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/misc.py
```diff
@@ -23,3 +25,11 @@ def as_choices(iterable):
...
     """
     return tuple((elem, elem) for elem in iterable)
+
+
+def random_pin_generator(count=4):
+    num = ""
+    for _ in range(1, count + 1):
+        n = random.randint(0, 9)
+        num += str(n)
+    return num
```
oudeismetis, line 35: doesn't matter much....but a 1-liner version
`return "".join([str(random.randint(0,9)) for _ in range(1, count + 1)])`

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -128,6 +134,38 @@ def reset_password(request, *args, **kwargs):
...
+
+    code_from_db = serializer.context.get("code_from_db")
+    code_from_db.is_used = True
+    code_from_db.save()
+
+    user.set_password(request.data.get("password"))
+    user.save()
+
+    u = UserLoginSerializer.login(user, request)
+    return Response(status=status.HTTP_200_OK, data=u)
```
oudeismetis, line 166: This is a lot of code to add for something that is only needed for mobile app builds, and we aren't able to remove it with a cookiecutter flag.
I'd like to discuss.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/templates/registration/password_reset_code.html
```diff
@@ -0,0 +1,41 @@
+{%- raw -%}
+{% extends '_action-email-base.html' %}
+{% load tz %}
```
paribaker, line 3: @oudeismetis not sure what to do about this file though

oudeismetis: files are easy actually.
Add logic to the post_gen_project hook to delete the file if users didn't pick the option.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -90,8 +93,37 @@ def test_password_reset(test_user, client):
...
+    code = 12345
+    UserResetPasswordCode.objects.create_code(user=test_user, code=12345)
+    res = client.post(
+        f"/api/password/reset/code/confirm/{test_user.email}/", {"code": code, "password": "testing12345"}, **JSON_RQST_HEADERS
+    )
+    assert res.status_code == 200
+    assert res.json()
+    res = client.post("/api/login/", {"email": test_user.email, "password": "testing12345"}, **JSON_RQST_HEADERS)
+    assert res.status_code == 200
+    assert res.json()
```
oudeismetis, line 123: Random aside. We need to get away from 200 checks.
Because of reasons, all responses will be a 200, so if you mistype the URL and are in a 404 condition, the status_code will still be 200.

So it's really the response payload that matters. I'm not sure but simply asserting that there is a json payload might be enough to avoid most issues we've seen with this.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/serializers.py
```diff
@@ -68,3 +74,21 @@ def validate(self, data):
...
         return User.objects.create_user(**validated_data)
+
+{%- if cookiecutter.include_mobile == 'y' %}
+class CodeResetPasswordSerializer(serializers.Serializer):
+    code = serializers.CharField(allow_blank=False, required=True)
+    password = serializers.CharField(allow_blank=False, required=True)
+
+    def validate_code(self, value):
+        code_from_db = self.context.get("user").reset_password_codes.filter(is_deactivated=False).first()
+        if not code_from_db or not code_from_db.is_valid or not check_password(str(value), code_from_db.code):
```
franagustin, line 85: Here we are first filtering for `is_deactivated=False` and then we check the `is_valid` which also checks that.

Maybe we could add a method in `UserResetPasswordCodeManager` to fetch only valid codes:

```py
class UserResetPasswordCodeManager(models.Manager):
    ...
    def filter_valid():
        return self.filter(is_deactivated=False, is_used=False)
    ...
```

Or maybe the `check_password` should also be in the `is_valid` method?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/signals.py
```diff
@@ -1,11 +1,37 @@
+import logging
+
+{%- if cookiecutter.include_mobile == 'y' %}
+from django.conf import settings
+{%- endif %}
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from rest_framework.authtoken.models import Token
 
+from {{ cookiecutter.project_slug }}.core.dispatch import new_reset_password_code_created_ds
```
franagustin, line 10: Should this be conditionally imported?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/signals.py
```diff
@@ -1,11 +1,37 @@
...
 from {{ cookiecutter.project_slug }}.core.models import User
+from {{ cookiecutter.project_slug }}.utils.emails import send_html_email
```
franagustin, line 12: Should this be conditionally imported?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/signals.py
```diff
@@ -1,11 +1,37 @@
...
+
+logger = logging.getLogger(__name__)
```
franagustin, line 14: Need to check but, if not mobile, will this give a linting error due to unused variable?
If so, we might also want to set it conditionally

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
franagustin: This is true and we may want to change that.
However, the default 404 page returned is not JSON and the `res.json()` would fail.
So I guess the test would work either way?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -128,6 +137,40 @@ def reset_password(request, *args, **kwargs):
     return Response(response_data, status=status.HTTP_200_OK)
 
 
+{%- if cookiecutter.include_mobile == 'y' %}
+@api_view(["get"])
```
franagustin, line 141: This view does actually create an object, I personally think that POST might be more adequate.
Any specific reason for the GET?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/templates/registration/password_reset_code.html
franagustin: Addressed!
https://github.com/thinknimble/tn-spa-bootstrapper/pull/192/commits/09d29a291604def39db4d7ec4bb1c12d0156d615

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
```diff
@@ -1,11 +1,15 @@
...
 from django.conf import settings
+{%- if cookiecutter.include_mobile == 'y' %}
+from django.contrib.auth.hashers import make_password
+{%- endif %}
 from django.contrib.auth.models import AbstractUser, BaseUserManager
 from django.contrib.auth.tokens import default_token_generator
 from django.db import models
 
 from {{ cookiecutter.project_slug }}.common.models import AbstractBaseModel
+from {{ cookiecutter.project_slug }}.core.dispatch import new_reset_password_code_created_ds
```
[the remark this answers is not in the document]
franagustin: Confirmed 😅 
https://github.com/thinknimble/tn-spa-bootstrapper/actions/runs/5979215672/job/16222847539?pr=192

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
paribaker: Yea @oudeismetis is right I think we moved away from doing that! I might have copied this from another project. 
@NC-Fran  The response would not be 404 if the route does not exist is what Ed is saying.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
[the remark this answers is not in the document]
paribaker: oops missed ! thanks

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/serializers.py
paribaker: actually the is valid is supposed to also check that it hasn't expired, I thought I had that in there but it appears to be missing :( either way you are right we can condense this into a single filter 

```
def filter_valid():
           return self.filter(is_deactivated=False, is_used=False, datetime_created__gte=timezone.now()-timedelta(minutes=x))
```

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
paribaker: nope no reason I just don't change things that already work, since we are using get in the webapp version I used the same here

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
franagustin: I think the response does not have a 404 status code, but still is a 404 page which when you try to convert into json, fails

oudeismetis: yea exactly, I assumed checking the json object would work.
I believe for SPA apps Google's crawler now inspects the page contents to see if it's displaying a 404 message.
Kinda feels like breaking the internet but 😐 

Either way...we can set a good example here in the bootstrapper with how we test and then collectively try and educate other devs on avoiding "200 only" assert statements on other projects.

paribaker:
Closing and moving to tn-gists I believe our current idea is that we should just use a webapp for forgot password as a first option!

PR #193, opened 2023-06-15
Feature: template preview endpoint

NC-Fran:
## What this does

Add an endpoint which renders templates based on the provided parameters.
This is useful, for example, for templates which we will be sending by email, preview allows a faster edition than having to setup emails and send one for each small change you want to see.

## Checklist
- [X] Add an endpoint which renders templates (as a generic, so it takes the name as a parameter)
- [X] Receive variables used in the context and provide them to the template
- [X] Allow nested variables (dicts inside dicts)
- [X] Allow for model variables (which need to be an instance of a Django model)

## How to test

GET or POST to http://localhost:8000/api/template_preview/
You MUST provide the template name either through a query parameter or in the body. i.e.: http://localhost:8000/api/template_preview/?template=core/index-placeholder.html

Any arguments of the context can be provided through query parameters or in the body.
Both multiform and JSON bodies should be handled (DRF does that).

You may use Insomnia or Postman if it is easier to add the parameters and it will still work.

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
@@ -1,11 +1,15 @@
+from unittest import mock
+
 import pytest
 from django.contrib.auth import authenticate
-from django.test import Client
+from django.test import Client, override_settings
 from pytest_factoryboy import register
+from rest_framework.response import Response
 
 from .factories import UserFactory
 from .models import User
 from .serializers import UserLoginSerializer
+from .views import PreviewTemplateView
 
 JSON_RQST_HEADERS = dict(
     content_type="application/json",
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -85,3 +89,53 @@ def test_password_reset(test_user, client):
...
+class TestPreviewTemplateView:
+    url = "/api/template_preview/"
+
+    @override_settings(DEBUG=False)
+    def test_disabled_if_not_debug(self, client):
+        response = client.post(self.url)
+        assert response.status_code == 404
+
+    @override_settings(DEBUG=True)
+    def test_enabled_if_debug(self, client):
```
paribaker, line 104: Why do this? Why does it matter if it is in Debug or not?

paribaker (approved):

paribaker:
I tried a get here: 
`https://tn-spa-bootstrapper-pr-193.herokuapp.com/api/template_preview/?template=core/index-placeholder.html/`
but did not see anything!

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
NC-Fran: I thought it would be wise to disable this view in production / production-like envs. It is only intended to ease development, so we do not actually need it in other environments

https://github.com/thinknimble/tn-spa-bootstrapper/pull/193/files#diff-8b5450c1f1b12638e18ae360a56f149bce98e774720b01f226cf1b8129a9f893R139-R140

NC-Fran:
> I tried a get here: `https://tn-spa-bootstrapper-pr-193.herokuapp.com/api/template_preview/?template=core/index-placeholder.html/` but did not see anything!

@paribaker Weird! I do not have access to this project in Heroku, but can you confirm settings.DEBUG is set to True in the review app? Otherwise, it will be just a non-existent url 😅

paribaker:
Ah I see this is only available in local, staging and review apps then. 
Looks great!

PR #194, opened 2023-06-16
Feature: Add an autocomplete list filter for foreign key in Django admin

NC-Fran:
## What this does

Django's default way of rendering admin list filters for foreign keys just sucks!
If your related model has many rows in the database, it can start slowing your admin site down really quickly, even making it unusable due to timeouts.

This PR intends to provide an easy-to-use out-of-the-box replacement: A select2 widget similar to the behaviour you can set for the detail change view if configured through **autocomplete_fields** attribute.
However, Django still does not provide a select2 widget for the filters. This solution implements a custom filter.

Changes this:

image

Into this:
image

## How to test

1. Use cookiecutter to create a project based on this branch: `cookiecutter https://github.com/thinknimble/tn-spa-bootstrapper.git --checkout feat/autocomplete-admin-list-filter`
2. Create a model which has a ForeignKey field, makemigrations and run them.
3. Create an Admin and set up a filter using AutocompleteFilter.
4.  Remember to make your Admin's Media class a subclass of AutocompleteAdminMedia.

## Code examples

**models.py**
```py
class MyModel(AbstractBaseModel):
    user = models.ForeignKey("User", on_delete=models.CASCADE)

    def __str__(self):
        return str(self.user)
```

**admin.py**
```py
from {PROJECT_NAME}.common.templates.common.filters import AutocompleteFilter, AutocompleteAdminMedia


@admin.register(MyModel)
class MyModelAdmin(admin.ModelAdmin):
    list_filter = (("user", AutocompleteFilter), )

    class Media(AutocompleteAdminMedia):
        pass
```

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/static/common/autocompleteFilter.js
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/static/common/autocompleteFilter.js
@@ -1,0 +1,23 @@
+'use strict';
+
+
+function applyFilter(event, parameterName) {
+  const prefix = getPrefixFromElement(event.target);
+  const value = document.getElementById(`${prefix}-${parameterName}`).value;
+  const newSearchParam = `${parameterName}=${value}`;
+  const stringToReplace = new RegExp(`([\\?|&])${parameterName}=[^&]*(&)?`);
+
+  if (stringToReplace.test(window.location.search)) {
+    // $1 will be the preceding question mark or ampersand and $2 the ampersand after if any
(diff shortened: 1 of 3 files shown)


paribaker:
This is great! Wonderful to see someone specialize in a framework so deeply! 
One comment I have is to move things to a much more specific directory, maybe templates/common/admin for the template and template related files. For the filter I would do the same move it to a folder called filters/admin/filters.pay
It's very common in our projects that someone will add api filters and it's best to differentiate what is for the admin and what is not.

✅🔥⭐️

oudeismetis (approved):

PR #200, opened 2023-06-27
Preserve old secrets when rerunning on existing projects

oudeismetis:
## What this does

Not a huge deal on the face of it, but you can re-run cookiecutter to update an existing project. If you do that though, it'll generate fresh secrets for you. The main issue with that is it will break your local dev database. So you then need to remember to ignore that change. 

This solves that.

## Checklist
- [x] Only generate random secrets if an existing project isn't detected
- [x] Refactor the logic a little

## How to test

Run cookiecutter locally using this branch. (dummy project)
git init, git add ., git commit -m "init"
Run it a second time.
Observe that there are no changes to commit with the 2nd run.

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -167,8 +167,8 @@ def set_keys_in_envs(django_secret, postgres_secret):
 
 def get_secrets():
     """
-        If updating an existing project, fetch the previously set secrets so there is no change
-        Otherwise, generate new random values
+    If updating an existing project, fetch the previously set secrets so there is no change
+    Otherwise, generate new random values
     """
     django_secret = get_random_secret_key()
     postgres_secret = get_random_secret_key()


hooks/post_gen_project.py
```diff
@@ -152,27 +148,47 @@ def remove_expo_yaml_files():
...
-    set_django_secret_key(pull_request_template_path, django_secret_key)
-    set_django_secret_key(cookie_cutter_settings_path, django_secret_key)
-    secret = get_random_secret_key()
-    set_flag(env_file_path, "!!!POSTGRES_PASSWORD!!!", value=secret)
-    set_flag(postgres_init_file, "!!!POSTGRES_PASSWORD!!!", value=secret)
+    set_flag(env_file_path, "!!!DJANGO_SECRET_KEY!!!", django_secret)
+    set_flag(pull_request_template_path, "!!!DJANGO_SECRET_KEY!!!", django_secret)
+    set_flag(cookie_cutter_settings_path, "!!!DJANGO_SECRET_KEY!!!", django_secret)
+    set_flag(env_file_path, "!!!POSTGRES_PASSWORD!!!", postgres_secret)
+    set_flag(postgres_init_file, "!!!POSTGRES_PASSWORD!!!", postgres_secret)
```
oudeismetis, line 160: minor refactor

hooks/post_gen_project.py
```diff
@@ -152,27 +148,47 @@ def remove_expo_yaml_files():
...
+        # Hacky...but users won't have to pip install a library for this to work on their machine
+        with open(".env.example.bak") as f:
+            for line in f:
+                line = line.strip()
+                if line.startswith("SECRET_KEY="):
+                    django_secret = line.removeprefix("SECRET_KEY=").strip().strip("'")
+                elif line.startswith("DB_PASS="):
+                    postgres_secret = line.removeprefix("DB_PASS=").strip().strip("'")
+        remove(".env.example.bak")
+    return django_secret, postgres_secret
```
oudeismetis, line 186: Tried doing this with a library.
Load the .env.example file in as config and just read the values.
Problem is that there is nothing in the standard library that does that. You have to bring in a 3rd party library.
Would rather not require users to pip install another package just so they can bootstrap.

hooks/pre_gen_project.py
```diff
@@ -1,11 +1,26 @@
...
+def main():
+    if hasattr(project_slug, "isidentifier") and not project_slug.isidentifier():
+        raise Exception(f"'{project_slug}' project slug is not a valid Python identifier.")
+    if project_slug != project_slug.lower():
+        raise Exception(f"'{project_slug}' project slug should be all lowercase")
+
+    # For rerunning against an existing project: Make a backup of the env example file.
+    # We'll use that to keep the existing secrets in place when post_get_project runs later
+    if exists(".env.example"):
+        copyfile(".env.example", ".env.example.bak")
```
oudeismetis, line 22: Refactor the make this file a little more official.

Also this is where we make a backup of the existing .env.example.

paribaker (approved):

PR #201, opened 2023-07-13
DX - Vue/React - Add prettier plugin for tailwind class standard ordering

lakardion:
## What this does
Add a prettier plugin to both React and Vue so that tailwind class names are sorted in a standard way.
Based on this: [Automatic Class Sorting with Prettier](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier)

## How to test

- Create a project with cookiecutter and this branch with either React or Vue as web clients
- Go to web client folder
- Install dependencies
- Try changing the className / class of a html element. A good class to test is: `text-white px-4 sm:px-8 py-2 sm:py-3 bg-sky-700 hover:bg-sky-800`
- Hit save (or trigger format the way you're used to)
- The class should get re-ordered


lakardion:
Okay not sure what is wrong with eslint vue... don't even know why would it care about prettier...

lakardion:
> Okay not sure what is wrong with eslint vue... don't even know why would it care about prettier...

Seems like vue has a  tough time with eslint 8+. Ended up just leaving the current version of things, wanted to do some upgrades but found these issues

paribaker (approved):
LGM 
A nice clean up

PR #202, opened 2023-07-14
DX - React/Native - Add Tanstack eslint plugin

lakardion:
## What this does
Add lint rule for best practices when using Tanstack Query.

https://tanstack.com/query/v4/docs/react/eslint/eslint-plugin-query


{{cookiecutter.project_slug}}/clients/web/react/src/utils/auth.tsx
```diff
@@ -89,7 +89,8 @@ export const useAuth = () => {
 export const useUser = () => {
   const { userId, updateToken, updateUserId } = useAuth()
   const navigate = useNavigate()
-  return useQuery(['user', userId], {
+  return useQuery({
+    queryKey: ['user', userId],
```
lakardion, line 93: IG: Here I changed this so that we comply to the lint rule. This was erroring due to that the recommended way of setting queries is through single parameter rather than multiple.

The lint rule also takes care of the key to have the right variables so that it doesn't run out of sync and refetches whenever any of those changes. 
IG: if we depend on a variable and we don't add it to the query key we would get an error:
image

paribaker (approved):
LGM !

PR #203, opened 2023-07-14
Web React - Upgrade tn-models, add zod

### LINKED ISSUE #199: Fix client dependencies
https://github.com/thinknimble/tn-spa-bootstrapper/pull/196/ reorganizes the client side so that it aligns part of the web & mobile structure of the apps. It looks like it may be missing `zod` as a dependency, so we should add that to the client's `package.json`, bootstrap a client project and check that everything runs out-of-the-box.

Bonus points:
 - review if there are no similar missing dependencies on the mobile side of the bootstrapper

lakardion:
## What this does

- Upgrade tn-models to latest
- Add missing dependency. Closes #199


paribaker (approved):
Nice!

PR #204, opened 2023-07-16
replace with new tn-models-fp

paribaker:
## What this does

Moves vue js to the new tn-models-fp library

## How to test
Sign up to app
login to app
Start forgot password flow (note emails are not set up)


{{cookiecutter.project_slug}}/clients/web/vue3/src/services/AxiosClient.js
```diff
@@ -18,7 +18,7 @@ class ApiService {
   static session
   static init
   constructor() {
-    let baseURL = `${window.location.protocol}//${window.location.host}/api`
+    let baseURL = `${window.location.protocol}//${window.location.host}/`
```
paribaker, line 22: Thoughts on this? We do not append the `api` to the domain in the react side, I think that is a better approach since it is more explicit and controlled by the dev

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/base-model.js
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod'
+import {readonly} from '@thinknimble/tn-models-fp'
+
```
paribaker, line 3: @lakardion getting a zod error here that id is required when using userApi.create(), I temporarily set it to `.optional` to get it working, but I thought that by making it readonly it would not be an issue?

lakardion: sounds like a bug. but at the same time you need not to make id readonly. it is readonly by default

lakardion: Tracking here. As a workaround, not marking `id` as readonly it should work!
https://github.com/thinknimble/tn-models-fp/issues/132

paribaker: okay will leave this as is for now

lakardion: I think there's a bug with the readonly implementation when used on `id`, due to its duality with being readonly regardless being marked as readonly or not. 
readonly should be fine for other fields, until I get to the aforementioned issue we should probably leave `id` without the `readonly` wrap

paribaker: Okay will do, regardless the issue was occurring before I added the readonly, I was just trying to fix it by using the readonly method

lakardion: Okay I didn't notice the create shape override. Yeah readonly is not supported in the override models. Will fix that so that they're completely ignored.

lakardion: Upgrade to `2.12.3` and lmk how it goes

https://github.com/thinknimble/tn-models-fp/pull/137

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
```diff
@@ -37,7 +38,7 @@ const resetPassword = createCustomServiceCall(
 export const userApi = createApi(
   {
     client: axiosInstance,
-    baseUri: '/api/users/',
+    baseUri: 'api/users/',
```
oudeismetis, line 41: fun

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/AxiosClient.js
oudeismetis: My thought would be the opposite. The AxiosClient is making all of it's calls to the backend API.
The baseURL for the backend API starts at `/api`.

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/models.js
```diff
@@ -1,14 +1,29 @@
-import Model, { fields } from '@thinknimble/tn-models'
+import { z } from 'zod'
+import { readonly } from '@thinknimble/tn-models-fp'
+import { baseModelShape,  } from '../base-model'
 
-import UserAPI from './api'
+export const userShape = {
+  ...baseModelShape,
```
oudeismetis, line 6: Why are these all "shapes"?

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
```diff
@@ -58,7 +58,7 @@ export default {
 
       const { uid, token } = route.params
 
-      User.api
+      userApi.csc
```
oudeismetis, line 61: what's `csc`?

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Signup.vue
```diff
@@ -92,8 +93,7 @@ export default {
       unwrappedForm.validate()
       if (!unwrappedForm.isValid) return
 
-      User.api
-        .registerUser({
+      userApi.create({
```
oudeismetis, line 96: `csc` missing here?

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
lakardion: Hmm I don't think that slash matters given that the axios instance should already have the baseUrl defined?, so whether or not you add the slash, axios will know whether it should add it or not

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/models.js
lakardion: https://github.com/thinknimble/tn-models-fp/issues/145

Basically they comply with type from `zod` library. 

We could call them `models` if we wanted to, but they still would need to match the `ZodRawShape` structure.

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
lakardion: It's just an alias for `customServiceCall`

https://github.com/thinknimble/tn-models-fp#createcustomservicecall

`.customServiceCall` can still be used if you preferred the longer name.

Ideally all method calls should probably be available at the root of the api object, but the typing for that was quite a pain so any custom call live under `customServiceCall`.

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Signup.vue
lakardion: These are just built-ins so they live at the root of the api object

https://github.com/thinknimble/tn-models-fp#built-in-methods

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
oudeismetis: I was just observing that when I've seen changes like this in the past, it's because things are fragile with slashes missing or extra ones sneaking in and it being easy for someone to break.
Ideally we have a URL builder or a utility that sanitizes URLs before we use them. This exists on the BE in Python as part of the standard library. Not sure if we have it on the frontend.

Not relevant for this PR. Just a code smell.

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/models.js
oudeismetis: gotcha. So we call them shapes because zod calls them shapes. Fair enough.
If it's an internal only zod thing that most consumers of zod wouldn't know about I would consider calling it something else.
Models may or may not be the answer there.

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/AxiosClient.js
mayakeeley: I'm with Ed on this one, I'd prefer /api stayed there

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
```diff
@@ -1,81 +1,48 @@
...
+    return utils.fromApi(res.data)
+  },
+)
+
+const requestPasswordReset = createCustomServiceCall(
+  {
+    inputShape: forgotPasswordShape,
+  },
+  async ({ client, input }) => {
+    console.log(input)
```
mayakeeley, line 22: Are we deliberately keeping a console.log here? Is there a reason for that?

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
```diff
@@ -22,6 +22,7 @@ const requestPasswordResetCode = createCustomServiceCall(
     await client.get(`api/password/reset/code/${input.email}/`)
```
mayakeeley, line 22: Not sure if I've misread this but it looks like all of the other api calls have been changed to remove code. Should this not match that pattern too?

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/AxiosClient.js
paribaker: sounds good will update appropriately

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
paribaker: Well actually @lakardion built tn-models to handle that (at least the  trailing slash) right?

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/models.js
paribaker: Yea I like the idea of shapes over models, but I could also be content with `structure` `entity` `interface` <--- maybe less this one. I think models is a bit incorrect since these aren't true db backed 'models'

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
lakardion: trailing is handled.
leading, imo doesn't matter. ran a quick test and the request is the same whether you do `api` or `/api` with axios

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
paribaker: ah this is a different endpoint to the one for the web views. In web we use the reset link, but in mobile we use a reset code. We found this approach to be a less cumbersome approach than trying to figure out and get `in-app` links working.

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
oudeismetis: nice!

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/models.js
oudeismetis: shapes seems fine then

{{cookiecutter.project_slug}}/clients/web/react/src/services/user/api.ts
```diff
@@ -14,12 +14,12 @@ const login = createCustomServiceCall(
...
 )
 
-const requestPasswordResetCode = createCustomServiceCall(
+const requestPasswordReset= createCustomServiceCall(
   {
     inputShape: forgotPasswordShape,
   },
   async ({ client, input }) => {
-    await client.get(`api/password/reset/code/${input.email}/`)
+    await client.get(`api/password/reset/${input.email}/`)
```
oudeismetis, line 22: a change here but none to the backend?

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
```diff
@@ -1,81 +1,47 @@
...
+import axiosInstance from '../AxiosClient'
+import { userShape, forgotPasswordShape, userCreateShape, loginShape } from './models'
+
+const login = createCustomServiceCall(
+  {
+    inputShape: loginShape,
+    outputShape: userShape,
+  },
+  async ({ client, input, utils }) => {
+    const res = await client.post('api/login/', utils.toApi(input))
```
oudeismetis, line 12: I think we undid that one change, but all these URLs still start with `/api`. That might be why Cypress is failing

{{cookiecutter.project_slug}}/clients/web/react/src/services/user/api.ts
paribaker: The backend is all set, we probably copied some of this code from an app that was using code only the mobile side needs this code!

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
paribaker: yep we made the change but I  thought we said it doesn't matter in the end!

paribaker:
@oudeismetis @lakardion okay this is giving me a headache. To resolve once and for all I've added the `/api` to the axios instance on all 3 client/mobile side apps and removed `api` or `/api` from all services folders!

Will test this out on mobile and then check the react side!

.github/workflows/pytest.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Pytest, McCabe, and Coverage
         run: |
           cd my_project
-          pipenv install --dev --deploy --skip-lock
+          pipenv install --dev --deploy
```
oudeismetis, line 48: why did we make this change?

oudeismetis:
> @oudeismetis @lakardion okay this is giving me a headache. To resolve once and for all I've added the `/api` to the axios instance on all 3 client/mobile side apps and removed `api` or `/api` from all services folders!
> 
> Will test this out on mobile and then check the react side!

I appreciate the efforts to keep things standard, though technically it doesn't matter if Vue and React differ a bit in their approach.
For our teams sake though, I'm sure it would be confusing to bounce between 2 projects doing things slightly differently like this.

.github/workflows/pytest.yml
oudeismetis: yeaaa....pip install couldn't possibly have the --dev flag. Lets merge this as-is and open a github issue to figure out a way to bring this back using pipenv

oudeismetis (approved):

PR #206, opened 2023-07-19
Version pin Cypress Github Action to 16

paribaker:
## What this does

Pins the node version for the cypress test on chrome to 16
Cypress tests are unstable without this as the Github Action container will use whatever is available. Sometimes that 16, sometimes it's 12. Pinning it forces it to use the one we need.


oudeismetis (approved):

oudeismetis:
Ignoring Heroku errors for now as they are unrelated

PR #212, opened 2023-08-10
Alwasy returning 404 from the server is a bug

oudeismetis:
## What this does

Merged this change because I thought it fixed API status code issues and believed that Cypress failing was a fluke.
I was wrong. 
Undoing the change.

## How to test

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
@@ -21,5 +21,5 @@ def index(request):
     try:
         return render(request, "index.html")
     except TemplateDoesNotExist:
-        return render(request, "core/index-placeholder.html")
+        return render(request, "core/index-placeholder.html", status=status.HTTP_404_NOT_FOUND)
 {% endif -%}
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
@@ -1,3 +1,5 @@
+import logging
+
 {% if cookiecutter.use_graphql == "y" -%}
 from django.template.response import TemplateResponse
 from django.views.decorators.cache import never_cache
@@ -8,6 +10,7 @@ from django.template.exceptions import TemplateDoesNotExist
 from rest_framework import status
 {% endif -%}
 
+logger = logging.getLogger(__name__)
 
 {% if cookiecutter.use_graphql == "y" %}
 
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
```diff
@@ -19,7 +19,7 @@ def index(request):
 
 def index(request):
     try:
-        return render(request, "index.html", status=status.HTTP_404_NOT_FOUND)
+        return render(request, "index.html")
     except TemplateDoesNotExist:
-        return render(request, "core/index-placeholder.html", status=status.HTTP_404_NOT_FOUND)
+        return render(request, "core/index-placeholder.html")
```
oudeismetis, line 24: @mayakeeley Something for next week.

This was the issue. Curious. I THOUGHT Vue would never hit this and so we would only get to this view on the Django side if Vue had a 404. But I guess I had that wrong.

I need to revisit what happens when you request a URL that Vue doesn't know about. Where does it send you next. That endpoint needs to raise a 404.
The reason for this fix was we'd have a problem where you would request a page, get the 404 page, but with a status code of 200. Fine for regular users, but when writing automated tests you could easily be tricked and tripped up by that.

mayakeeley: I can see how removing it on line 22 might make a difference but I would've thought line 24 would be fine

oudeismetis: OK, I think I originally tested this on a project that was bootstrapped before the 404 route was added in 2022.
https://github.com/thinknimble/tn-spa-bootstrapper/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/clients/web/vue3/src/router/index.js#L46

So Vue catches the 404 situations. As a result, they HAVE to be `200` success. I tried looking for a workaround at the Vue level but couldn't find one.

The alternative would be to remove this, have it miss on the Vue level, and hit the server. The server would then serve a 404 page with a 404 HTTP status. Negatives here would be a non-Vue controlled page and the fact that it would be an extra server request.

The negative with reverting this logic and keeping it as it was is dev confusion. A dev debugging will see 200 reponses and assume a request succeeded, even though they mistyped the URL and it was actually a 404. (Has happened often in automated tests, especially the lazy ones where the contents of the response aren't inspected)

oudeismetis: https://stackoverflow.com/questions/54218371/how-to-get-a-404-response-in-vue-router

and 

https://stackoverflow.com/questions/73098868/vuejs-real-seo-friendly-404-page-for-search-engines-like-google-msn

for some interesting context

oudeismetis: I don't really want to send to the server as the fallback as it means we can't easily build fancy 404 pages that do things like recommend resources similar to what you were trying to find.

So the solution here might (unfortunately) have to be a dev education one. "Know that SPAs always return 200, don't trust it, make sure you validate the content returned."

paribaker (approved):

oudeismetis:
Step 1 is to undo the change.
Future improvements can be handled in a separate PR.

PR #218, opened 2023-08-24
Clients - Upgrade tn models, move to @thinknimble/tn-models 2.0.0

lakardion:
## What this does

- Upgrade tn-models to latest in all clients.
- Replace tn-models-fp with v2 of tn-models

tn-models-fp has been deprecated in favor of tn-models. Now any new updates to the packages will be deployed to `@thinknimble/tn-models`


paribaker (dismissed):

paribaker (approved):

paribaker:
this build was working with no errors, but is now testing the react side against the vue build. Will merge!

PR #229, opened 2023-09-08
fix: update server docker python version to match the one in the Pipfile

franagustin:
## What this does
Just updates Python version used by Docker to match the one currently in the Pipfile.
As of now, it is 3.10 instead of 3.7

## Checklist
- [X] Change version in Dockerfile

## How to test

Create a new project: `cookiecutter git@github.com:thinknimble/tn-spa-bootstrapper.git`
Run `pipenv install` and `docker compose up server`.
You will get an error due to which packages can't be installed.

Create a new project based off this branch: `cookiecutter git@github.com:thinknimble/tn-spa-bootstrapper.git -checkout fix/update-server-docker-python-version`
Run `pipenv install` and `docker compose up server`.
You will see successful logs and be able to access the admin site at `localhost:8000/staff/`


paribaker (approved):
LGTM

PR #232, opened 2023-09-15
feat: add examples on how to use the admin autocomplete filter

franagustin:
## What this does

Add some examples of how to use the autocomplete admin filter implemented by [#194](https://github.com/thinknimble/tn-spa-bootstrapper/pull/194)

## Checklist
- [X] Add an autocomplete filter for groups on the user admin list
- [X] Add an autocomplete filter for permissions on the group admin list

## How to test

1. Create a new project based off this branch: `cookiecutter git@github.com:thinknimble/tn-spa-bootstrapper.git -checkout feat/admin-autocomplete-filter-examples`
2. Run `pipenv install` and `docker compose up server`.
3. Go to the Group's admin page: http://localhost:8000/staff/auth/group/ . You should be able to see a filter for permissions which uses this widget.
4. Create a Group
5. Go to the User's admin page: http://localhost:8000/staff/core/user/ . You should be able to see a filter for groups which uses this widget.
**NOTE**: The filter will only appear if there is at least one instance of the related model in the DB, hence why creating a Group is necessary!


paribaker (approved):
LGTM

PR #243, opened 2023-10-06
fix import not using template slug

r2dliu:
## What this does

fixes an import not using the slug

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


franagustin (approved):
Oopsie, my bad! Good catch!

PR #248, opened 2023-11-08
some clean up for components and addition of loading spinner

paribaker:
## What this does

Cleans up some of the warnings from the components, also adds an additional component for loading and loading spinner

credits: @mayakeeley @bwennuh 

## How to test

Try to login with an incorrect login to see it


{{cookiecutter.project_slug}}/clients/web/vue3/src/components/inputs/InputField.vue
```diff
@@ -29,7 +30,7 @@
 </template>
 
 <script>
-import { computed } from 'vue'
+import { useModelWrapper } from '@/composables/VModelWrapper'
```
paribaker, line 33: This is a prebuilt util/composable

{{cookiecutter.project_slug}}/clients/web/vue3/src/components/inputs/InputField.vue
```diff
@@ -53,14 +54,14 @@ export default {
...
   },
   emits: ['blur', 'focus', 'input', 'update:value'],
-  setup(props, context) {
-    const val = computed({
-      get: () => props.value,
-      set: (value) => context.emit('update:value', value),
-    })
-
+  setup(props, { emit }) {
+    const val = useModelWrapper(props, emit, 'value')
```
paribaker, line 64: Putting this in as an example so that folks can see how it works!

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Login.vue
```diff
@@ -20,9 +20,6 @@
         />
         <div>
           <div class="flex items-center justify-between">
-            <label for="password" class="block text-sm font-medium leading-6 text-primary"
```
paribaker, line 23: This was unnecessary since it is part of the component and without it the component adds a undefined

mayakeeley: Fair but it was in here so that Password and Forgot Password could sit side by side -- I think it's fine like this but forgot password should sit on the rhs

mayakeeley (commented):
looks good aside from the forgot password thing!

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Login.vue
paribaker: Ah I see the only reason I moved it was because of the weird `undefined-label`
Would it be better located under create and account? or the RHS of that

paribaker: Ahh also that means the label for wasnt working!

paribaker: Done I made it into a slot!

mayakeeley (approved):
lgtm!

PR #249, opened 2023-11-10
Set up JS monorepo (Lib + Mobile + Web)

lakardion:
# What this does
- [x] Add more readable prompts for cookiecutter (needs user to have cookiecutter > 2.1.1 , does not break if used with older versions)
- [x] Add services core base library project
- [x] Allow services core to be an option in cookiecutter prompts
- [ ] Add cookiecutter services core reference to the current services uses in both web clients and the mobile client
  - [x] React
  - [x] React Native
  - [ ] Vue   

# What’s this for

Allow clients to share service layer code.

# Reasoning

Having two clients (mobile and web) usually results in code repetition on the service layer. This leads to the idea of extracting this code to a library that could be shared across them.

# Principle

This configuration uses npm workspaces which condenses the package management to the root of the project.
This in favor of reusing package references across the different workspaces in case any is used in more than one. 
Packages that are solely used in a single workspace are kept within the workspace’s node_modules instead

# Extra

## Caveats
- ‼ Adds a lot of cookiecutter conditions to files which will make them harder to read.
- ⚠️ Not tested with Docker (most likely won't work)
- Need to do a `npm install --workspace=services-core` before anything else to get the general install to work properly (Post installs weren't working otherwise)

### Allow the library to be aware of the current backend url

The current version of this setup exports callback services. In the sense that they still need to be provided an axios instance to be useable.
A good improvement to this would be to somehow inject the backend url into the service build step so that we can at least create the base axios instance that will be used in all services.

This way we wouldn’t need to use callbacks for the services in the clients, also this would lift the axios module requirement to the library rather than in the clients

### Figure out how to split the library in submodules

Right now everything from the library is exported from its root. We gave a couple of quick attempts to define multiple exports from the library such as `import from 'library/submodule'` but were unsuccessful, we might have missed some required configuration.


cookiecutter.json
```diff
@@ -27,5 +27,16 @@
...
+  "include_mobile": "n",
+  "include_services_core": "n",
+  "__prompts__": {
+    "project_name": "Project name:",
+    "author_name": "Author:",
+    "email": "Email:",
+    "project_slug": "The project slug",
+    "mail_service": "Choose an email service to use",
+    "client_app": "Choose a web frontend client",
+    "include_mobile": "Want to add a mobile project? (y/n)",
```
paribaker, line 39: 💯

lakardion:
Will move this to a guide instead due to my comment here:
```
Caveats
‼ Adds a lot of cookiecutter conditions to files which will make them harder to read.
⚠️ Not tested with Docker (most likely won't work)
Need to do a npm install --workspace=services-core before anything else to get the general install to work properly (Post installs weren't working otherwise)

```

PR #255, opened 2023-12-27
React - Match styles to vue client

noriega2112:
## What this does

Styling enhancement that matches vue client styles


lakardion:
Forgot and reset password were added here #276 
So you could probably continue with the styling here after the merge

lakardion:
@noriega2112 Do you remember what's missing this to be ready?

{{cookiecutter.project_slug}}/clients/web/react/src/styles/buttons.css
lakardion, line 1: This beats the purpose of tailwind 😅 but okay. This is how it is done in vue right?
We added tailwind so that we didn't need css files but we're coming back to the old days of creating css and going through the trouble of giving names to things

I mean, what's the difference with having this DIRECTLY in the button component than here? Why do we need this to be global at all ? Tailwind is here so that we don't have think about class names and have everything in the same place for colocation reasons.
I would propose removing this and leaving it in the button component instead

{{cookiecutter.project_slug}}/clients/web/react/src/styles/headings.css
lakardion, line 1: Again 🫠 I'd push back on using this. No need at all to have to define class names for these stuff.
If we want headings sub headings titles and stuff why don't we just create components? Plus this is very specific for each project, we may not need all these classes

{{cookiecutter.project_slug}}/clients/web/react/src/styles/index.css
lakardion, line 1: can't this live in a component?

{{cookiecutter.project_slug}}/clients/web/react/src/styles/nav-bar.css
lakardion, line 1: Still don't get the point why this has to be so far from the component that uses this and masked with class names

{{cookiecutter.project_slug}}/clients/web/react/src/utils/routes.tsx
```diff
@@ -1,37 +1,43 @@
-import { Navigate, Route, Routes } from 'react-router-dom'
-import { Home, LogIn, SignUp } from 'src/pages'
-import { ForgotPassword } from 'src/pages/forgot-password'
+import React from 'react'
```
lakardion, line 1: we can avoid the import if we just do `<></>`

lakardion (commented):
I would not like the react side to have custom tailwind classes. I'd rather have those styles to be specifically defined wherever the styles are going to be applied. 

That's one of the reasons why tailwind is so good. Co-locating style wherever it is being used. 

I don't see much point in abstracting and creating names for things that are going to be used in a single place. This hides the styles from devs. Yes, it takes a search but wouldn't it be better if we didn't even need that at all?
I think all these classes here are not needed at all and those styles could be placed (much) closer to where they're being used

lakardion:
Other than that running it looks good. I just fixed an issue with the templating in home.tsx

paribaker (commented):
Looks like there is just one small bug, the default route should be Home, right now it shows a blank page!

lakardion (approved):

paribaker (approved):

PR #256, opened 2024-01-03
Higher API security by default and improve logs

oudeismetis:
## What this does

API endpoints should require user login by default. If an endpoint is allowed to be opened to all, then it should be explicitly opened up. 

## Checklist
- [x] Require user auth by default for all API endpoints
- [x] Prevent duplicate logs
- [x] Handle logging JSON objects better

## How to test


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -158,7 +159,7 @@
 REST_FRAMEWORK = {
     "DEFAULT_VERSIONING_CLASS": "rest_framework.versioning.AcceptHeaderVersioning",
     "DEFAULT_PERMISSION_CLASSES": [
-        "rest_framework.permissions.AllowAny",
+        "rest_framework.permissions.IsAuthenticated",
```
oudeismetis, line 162: By default, all API endpoints will require that the user is authenticated. 
You can add the `AllowAny` permission to the permission classes for any view that needs it.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -308,16 +309,25 @@
 #
 # Custom logging configuration
 #
+
+
+class MyFilter(logging.Filter):
+    def filter(self, record):
+        record.msg = str(record.msg).replace("'", '"')
+        return True
```
oudeismetis, line 317: Per the docs, filters are used to determine if a given event should or should not be logged. But the docs also say you can use them to do things like this.

Our JSON objects weren't logging correctly because they were using single quotes instead of double quotes.
With this change, we'll now get more powerful papertrail logs like this:

image

oudeismetis: not that it's collapsed by default but theres a little button there to expand/collapse the JSON object.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -329,23 +339,20 @@
...
         },
         # The logger name matters -- it MUST match the name of the app
         "{{ cookiecutter.project_slug }}": {
-            "handlers": [
-                "console",
-            ],
+            "handlers": ["console"],
             "level": "DEBUG",
-            "propagate": True,
+            "propagate": False,
```
oudeismetis, line 355: This solves the duplicate logs issue. When you propagate, the log event also goes up to the parent logger. If that logger also logs it, then you'll end up with 2 events in a row every time.

paribaker (approved):

PR #262, opened 2024-01-09
updated react to follow new cypress as is used in vue

paribaker:
## What this does

This PR updates the react side to use the new cypress config set up that the Vue3 app uses. 

@oudeismetis This appears to be failing because it is building the vue instead of react. 

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/test-login.cy.ts
```diff
@@ -2,12 +2,9 @@ describe('Tests login workflow', () => {
...
     cy.visit('/')
     cy.url().should('include', '/log-in')
-  })
-  it('Filling in email and passwords goes to', () => {
     cy.get('[data-cy=email]').type(Cypress.env('TEST_USER_EMAIL'))
     cy.get('[data-cy=password]').type(Cypress.env('TEST_USER_PASS'))
-    cy.get('[data-cy="login-btn"]').click()
-    cy.url().should('include', '/home')
+    cy.contains('[data-cy="login-btn"]', 'Login').click()
+    cy.url().should('include', '/test')
```
oudeismetis, line 8: To keep consistent with the vue side we should keep this as 2 separate tests.
The first test is nice as if it fails it usually means the app failed to build/load all together. But if it passes and the second one fails, it is usually a BE/data/config issue.

and of course...we need more/better tests here anyway. But that's a separate issue.

oudeismetis:
I rebased, thinking that would trigger a new build and fix the issue, but it didn't. If you push again it should do it.

{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/test-login.cy.ts
paribaker: Ok for simplicity can we approve this and I'll make the changes on a project I am working on. The reason I included it all here is because it fails as two separate ones, see this migration guide 
[here](https://docs.cypress.io/guides/references/migration-guide)

lakardion (commented):
lgtm here I'll make some time to give it a go locally

paribaker:
> I rebased, thinking that would trigger a new build and fix the issue, but it didn't. If you push again it should do it.

Is there something I am doing incorrectly that would make it not build the react?

oudeismetis (dismissed):

{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/test-login.cy.ts
```diff
@@ -2,12 +2,9 @@ describe('Tests login workflow', () => {
   it('Home page auto redirects to login', () => {
     cy.visit('/')
     cy.url().should('include', '/log-in')
-  })
-  it('Filling in email and passwords goes to', () => {
     cy.get('[data-cy=email]').type(Cypress.env('TEST_USER_EMAIL'))
     cy.get('[data-cy=password]').type(Cypress.env('TEST_USER_PASS'))
-    cy.get('[data-cy="login-btn"]').click()
-    cy.url().should('include', '/home')
+    cy.contains('[data-cy="login-btn"]', 'Login').click()
```
lakardion, line 7: As you now are also checking for the button content. This should be `Log in`. Right now the test is failing bc of this

{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/test-login.cy.ts
lakardion, line 8: Should be `/home`?

lakardion (approved):
Did a local run. These were my findings!

oudeismetis (approved):

PR #263, opened 2024-01-09
Fail CI checks when prettier has warnings

oudeismetis:
## What this does

By default CI/CD should error for prettier style issues.


{{cookiecutter.project_slug}}/clients/web/vue3/.eslintrc.cjs
```diff
@@ -35,6 +35,6 @@ module.exports = {
     'vue/no-v-model-argument': process.env.NODE_ENV ? 'warn' : 'off',
     'vue/multi-word-component-names': process.env.NODE_ENV ? 'warn' : 'off',
     '@typescript-eslint/no-unused-vars': 'warn',
-    'prettier/prettier': 'off',
+    'prettier/prettier': 'error',
```
lakardion, line 38: @oudeismetis 
Ah I didnt see this 🤦 .
No, this is ESLint getting in the way of prettier, this is why we're using a plugin to properly separate ESLint from prettier so they don't conflict. 
This Rule is quite annoying as it tells you live that there are formatting errors when you're not even finishing writing.

This check can go in the PR action rather. I just did that here #272

PR #273, opened 2024-01-26
Clients: Fix - CSRF Token issues

lakardion:
## What this does

- Add call to `/logout` to clear cookies in browser (added to mobile as well but  I don't think that changes anything there)

## What fixes
We had this issue for some time where if you 
Log In > Log Out > Sign up
There was a `sessionid` cookie that didn't get cleaned up on logout. This cookie was httpOnly which didn't allow the FE js to access and delete it. This needed to be done by the backend so for that we needed to call the `/logout` endpoint which is what we're doing in this PR.

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/web/vue3/src/components/NavBar.vue
+++ b/{{cookiecutter.project_slug}}/clients/web/vue3/src/components/NavBar.vue
@@ -131,6 +131,7 @@
 </template>
 
 <script>
+import { userApi } from '@/services/users'
 import { computed, ref } from 'vue'
 
 import { useRouter } from 'vue-router'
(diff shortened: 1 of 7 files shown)


{{cookiecutter.project_slug}}/clients/web/vue3/src/components/NavBar.vue
```diff
@@ -143,7 +144,8 @@ export default {
     let mobileMenuOpen = ref(false)
     let profileMenuOpen = ref(false)
 
-    function logout() {
+    async function logout() {
+      await userApi.csc.logout()
```
lakardion, line 148: wonder whether this actually works in Vue

lakardion: the call was made, but well not sure whether this is legal

oudeismetis (approved):

PR #275, opened 2024-01-29
Hotfix - Clients - React wrong import of function crashing on fresh bootstraper

lakardion:
## What this does

Fix wrong import in log-in.

## How to test


oudeismetis (approved):

oudeismetis:
It bootstrapped the wrong type of app which is why cypress is failing.
But I have noticed Cypress failing on a newly bootstrapped React webapp. Not sure why.

PR #276, opened 2024-01-30
React Client - Rework Auth routes, improve user api service

lakardion:
- add reset pass
- add forgot pass
- rework user api to properly reflect models and the right custom calls. Rm token from user default model which only belongs to signup and login
- rework routes to stop using app or auth component util

## What this does

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


oudeismetis (commented):
This is a test review to trigger a Webhook

paribaker:
@lakardion can we also add these two 

```
export const BounceableWind = styled(Bounceable, {
  props: {
    contentContainerStyle: true,
  },
})

export const KeyboardAvoidingViewWind = styled(KeyboardAvoidingView, {
  props: {
    style: true,
    contentContainerStyle: true,
  },
})
```

paribaker (approved):
👍

Issue #242, opened 2023-10-06
Remove IN_REVIEW from aws storage settings


paribaker:
Since this is managed with the USE_AWS_STORAGE var, this one should allow users to set it if they want!

whusterj:
Stale.

PR #277, opened 2024-01-31
Remove IS_REVIEW_APP from aws config

eliasbiagioninc:
## What this does

PR for issue https://github.com/thinknimble/tn-spa-bootstrapper/issues/242 

## Checklist
- [x] Remove IS_REVIEW_APP from aws config in settings.py

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
@@ -260,7 +260,7 @@ else:
 
 PRIVATE_MEDIAFILES_LOCATION = ""
 # Django Storages configuration
-if config("USE_AWS_STORAGE", cast=bool, default=False) and not IS_REVIEW_APP:
+if config("USE_AWS_STORAGE", cast=bool, default=False):
     AWS_ACCESS_KEY_ID = config("AWS_ACCESS_KEY_ID")
     AWS_STORAGE_BUCKET_NAME = config("AWS_STORAGE_BUCKET_NAME")
     AWS_SECRET_ACCESS_KEY = config("AWS_SECRET_ACCESS_KEY")


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -260,7 +260,7 @@
 
 PRIVATE_MEDIAFILES_LOCATION = ""
 # Django Storages configuration
-if config("USE_AWS_STORAGE", cast=bool, default=False) and not IS_REVIEW_APP:
+if config("USE_AWS_STORAGE", cast=bool, default=False):
```
oudeismetis, line 263: Not necessarily opposed to the removal, but a couple of things:

- See https://github.com/thinknimble/tn-spa-bootstrapper/pull/138 as @paribaker added a reason for this change being made
- If we are going to remove this, then perhaps we consider removing `IS_REVIEW_APP` and `IN_REVIEW` as well, since they have very limited use

paribaker: well the reason it is in there is so that we do not by default store things and instead explicitly do it (but I guess you can also just not provide the .env variables). The reason for this is because when the review app is destroyed we do not have a process for clearing out the assets

oudeismetis: yea, and I know some have run into this "annoying" extra safe guard we added when trying to do something specific in a review app. Just want to make sure Elias is thinking of the implications and any broader changes.

I personally think both `IS_REVIEW_APP` and `IN_REVIEW` should just be removed all together. But I'm also fairly indifferent on this as well.

paribaker: If it becomes an issue we can re-add it ! Approved for a merge IMO! @eliasbiagioninc

paribaker (approved):

PR #282, opened 2024-02-14
Clients - React Native - Fix nativewind types after recent change in TS

lakardion:
## What this does

https://github.com/marklawlor/nativewind/pull/731

## How to test


{{cookiecutter.project_slug}}/clients/mobile/react-native/global.d.ts
lakardion, line 1: the actual fix is the renaming of this file

paribaker (approved):

PR #286, opened 2024-03-01
Refactor - Mobile - Remove irrelevant services context

lakardion:
## What this does

- remove Services provider context not being useful and just adding an extra layer to get to navio stable value (doens't change so it doesn't need a context) and in case we need it to change we would probably just use a better alternative than context instead.
- Remove unused translation service. In case we need this in a service we could make it atomic some time in the future.


{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/index.tsx
lakardion, line 1: This wasn't doing anything really. Navio instance doesn't change so it is the same if we just use the getNavio call or this hook.
The translation service wasn't even being used in most projects. There's also the chance to move these to a better alternative than context. Either a Zustand store or atomic state with jotai

paribaker (approved):

PR #293, opened 2024-03-22
update swagger implementation

paribaker:
## What this does

We are working on some tn-models automations and would be helpful to have this added!

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


oudeismetis (approved):
LGTM @OkayItsMikael I don't think this is the library you were looking at. I assume this is compatible with the research you were doing. If not, would be easy to migrate later since it's low impact code wise.

paribaker:
> LGTM @OkayItsMikael I don't think this is the library you were looking at. I assume this is compatible with the research you were doing. If not, would be easy to migrate later since it's low impact code wise.

We pulled this out of the PR @OkayItsMikael had in the presentation

PR #295, opened 2024-03-23
new way of env vars

paribaker:
## What this does

Since expo SDK 49 we can bring env vars in a uniform manner using the default eas.json, we can also override these variables with variables from the .env.example (eg when working against a local ngrok backend) by prefixing the env variable with `EXPO_PUBLIC_`

Edit (damian)
Source:
[Reading environment variables from .env files](https://docs.expo.dev/guides/environment-variables/#reading-environment-variables-from-env-files)

This is convenient :
> Variables can be updated as you edit your code without restarting the Expo CLI

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


lakardion (approved):
lgtm

Added some source information in the description in case someone wants to take a deeper look on how this works

PR #296, opened 2024-03-25
Fix test settings so the built in `test_password_reset` passes

whusterj:
## What this does

I've had to fix this on a couple projects. The built-in unit test `test_password_reset` fails locally because Django whitenoise `CompressedManifestStaticFilesStorage` expects logo.png to exist in the manifest. But that only works after `collectstatic` has been called, making this test flaky.

This PR adjusts the `STATICFILES_STORAGE` value in `test_settings.py` so that the whitenoise manifest check is bypassed and so `test_password_reset` will pass.

## How to test

Bootstrap a new project and run unit tests locally. They no longer fail.

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/test_settings.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/test_settings.py
@@ -2,6 +2,10 @@ from decouple import config
 
 from {{ cookiecutter.project_slug }}.settings import *  # noqa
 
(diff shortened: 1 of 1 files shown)


paribaker (approved):

whusterj:
I've had to fix this on a couple projects. The built-in unit test test_password_reset fails locally because Django whitenoise CompressedManifestStaticFilesStorage expects logo.png to exist in the manifest. But that only works after collectstatic has been called, making this test flaky.

This PR adjusts the STATICFILES_STORAGE value in test_settings.py so that the whitenoise manifest check is bypassed and so test_password_reset will pass.

PR #298, opened 2024-03-26
Clients - React native upgrades - Expo, sentry-expo deprecation, NativeWind v4

lakardion:
## What this does

- Upgrade to [expo latest](https://expo.dev/changelog/2024/01-18-sdk-50)
- [Deprecate sentry-expo in favo rof @sentry/react-native](https://github.com/expo/fyi/blob/main/sentry-expo-migration.md)
- [Add changes for Upgrading to Nativewind v4](https://www.nativewind.dev/v4/getting-started/expo-router)

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/clients/mobile/react-native/babel.config.js
```diff
@@ -1,25 +1,6 @@
...
-            '@assets': './assets',
-            '@services': './src/services',
-            '@utils': './src/utils',
-            '@src': './src',
-          },
-          extensions: ['.js', '.jsx', '.ts', '.tsx'],
-        },
-      ],
-    ],
+    presets: [['babel-preset-expo', { jsxImportSource: 'nativewind' }], 'nativewind/babel'],
```
lakardion, line 4: Expo 50 now simply reads the `tsconfig` file and makes this without the need of `module-resolver`

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/components/styled.tsx
```diff
@@ -1,22 +1,9 @@
...
-import { KeyboardAvoidingView, ScrollView } from 'react-native'
+import { cssInterop } from 'nativewind'
+import type { FC } from 'react'
 import { Bounceable } from 'rn-bounceable'
 
-export const ScrollViewWind = styled(ScrollView, {
-  props: {
-    contentContainerStyle: true,
-  },
+cssInterop(Bounceable, {
```
lakardion, line 5: New way of doing NativeWind components.
We register the component, mark the prop that's going to be passthrough and then for Typescript purposes we cast the component to show up the new prop.

For native components (such as FlatList, ScrollView) these props are already taken cared of so you can find `contentContainerClassName` in `ScrollView` without the need of declaring a `ScrollViewWind` how we were doing previously

paribaker (commented):
FYI this is failing because of this file 

https://github.com/thinknimble/tn-spa-bootstrapper/blob/feature/expo-upgrade-50/resources/app-bs.config.js

Maybe we can start auto generating it rather than hardcoding

paribaker:
@lakardion lint will fail until merged with main so I think this is ready

paribaker:
Even though not all checks are passing, I will go ahead and merge this. The checks are not passing because there was a change in the package.json and the linting YAML file that will not take effect as it

PR #299, opened 2024-04-05
[PROTO] - Check bun as a package manager

lakardion:
## What this does

Check `bun` as a package manager for clients

Will try out this buildpack: https://github.com/jakeg/heroku-buildpack-bun

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


lakardion:
DROP because of cypress not being supported by Bun. We can give another try to this if we move to playwright at some point

ref: https://github.com/cypress-io/cypress/issues/28962

PR #304, opened 2024-05-13
Vue: Password reset broken

oudeismetis:
## What this does

This is not the first time this has broken.
Adding tests to help avoid this from happening again.

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
+++ b/{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
@@ -1,7 +1,7 @@
 import { createApi, createCustomServiceCall } from '@thinknimble/tn-models'
 import { z } from 'zod'
 import axiosInstance from '../AxiosClient'
-import { userShape, forgotPasswordShape, userCreateShape, loginShape } from './models'
+import { userShape, forgotPasswordShape, resetPasswordShape, userCreateShape, loginShape } from './models'
 
 const login = createCustomServiceCall(
   {
@@ -18,18 +18,22 @@ const requestPasswordReset = createCustomServiceCall(
   {
     inputShape: forgotPasswordShape,
   },
-  async ({ client, input }) => {
-    await client.get(`/password/reset/${input.email}/`)
+  async ({ client, input, utils }) => {
+    const res = await client.post('/password/reset/', utils.toApi(input))
+    return utils.fromApi(res.data)
   },
 )
+
 const resetPassword = createCustomServiceCall(
   {
-    inputShape: { uid: z.string().email(), token: z.string(), password: z.string() },
+    inputShape: resetPasswordShape,
     outputShape: userShape,
   },
   async ({ client, input, utils }) => {
-    const { email, ...rest } = utils.toApi(input)
-    const res = await client.post(`/password/reset/code/confirm/${email}/`, rest)
+    const { password } = utils.toApi(input)
+    const res = await client.post(`/password/reset/confirm/${input.uid}/${input.token}/`, {
+      password,
+    })
     return utils.fromApi(res.data)
   },
 )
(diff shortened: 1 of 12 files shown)


{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.js
```diff
@@ -18,18 +23,22 @@ const requestPasswordReset = createCustomServiceCall(
...
+    inputShape: resetPasswordShape,
     outputShape: userShape,
   },
   async ({ client, input, utils }) => {
-    const { email, ...rest } = utils.toApi(input)
-    const res = await client.post(`/password/reset/code/confirm/${input.email}/`, rest)
+    const { password } = utils.toApi(input)
+    const res = await client.post(`/password/reset/confirm/${input.uid}/${input.token}/`, {
+      password,
+    })
```
oudeismetis, line 41: This was fixed a while ago but somehow got re-broken in the past couple of months.

{{cookiecutter.project_slug}}/clients/web/vue3/tests/e2e/specs/test-password-reset.cy.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  it('Submitting email doesnt throw error', () => {
+    cy.visit('/login')
+    cy.contains('[data-cy=password-reset]', 'Forgot password?').click()
+    cy.url().should('include', '/password/request-reset/')
+    cy.get('[data-cy=email]').type(Cypress.env('TEST_USER_EMAIL'))
+    cy.contains('[data-cy=submit]', 'Request Password Reset').click()
+    cy.get('[data-cy=submit-success]').should('contain.text', 'Your request has been submitted')
+  })
+})
```
oudeismetis, line 15: Adding a basic test here to help avoid this breaking again.
If the backend has an error, the frontend won't show the `"Your request has been submitted..."` message

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/conftest.py
```diff
@@ -1,8 +1,25 @@
...
+    @pytest.mark.use_requests
+    """
+    if "use_requests" not in request.keywords:
+        exception = Exception("Do not call the internet from a unit test")
+        with mock.patch("requests.get", side_effect=exception), mock.patch("requests.post", side_effect=exception):
+            yield
+    else:
+        yield
+
+
```
oudeismetis, line 22: Bringing in this mock as well. The main goal is to force devs to think before they intentionally decide to write a test that will call the internet.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -71,12 +72,19 @@ def test_user_can_login(test_user):
...
+    post_request = rf.post("api/password/reset/", {"email": test_user.email})
+    request_reset_link(post_request)
+
+    # Grab from the logs the actual URL link we would send to the user
+    password_reset_creds = caplog.text.split("password/reset/confirm/")[1].split('"')[0]
+    password_reset_url = f"/api/password/reset/confirm/{password_reset_creds}/"
+
+    # Verify the link works for reseting the password
     response = client.post(password_reset_url, data={"password": "new_password"}, format="json")
     assert response.status_code == 200
```
oudeismetis, line 89: Making this test smarter to catch more password reset bugs. When email sending isn't configured, the email contents are sent to logs for dev convenience.
This test looks at those logs to confirm that the actual URL we generate is a valid one.

oudeismetis: Also notice the use of the `use_requests` marker here.
When building the HTML email we inline the CSS. The library we use to do that has to call the internet to get the styles.

I think the use of this marker could be treated like a code smell. Allow it for the moment, but ideally we would come back and properly mock that call to the internet to improve speed and stability of the test suite.

lakardion (approved):

PR #307, opened 2024-05-30
Vue: Tanstack and typescript

paribaker:
## What this does

Introduces Tanstack as a query client
Switches out the User services to use Typescript
Moves the user functions to a common hook


{{cookiecutter.project_slug}}/clients/web/vue3/src/components/NavBar.vue
```diff
@@ -145,11 +145,19 @@ export default {
     let profileMenuOpen = ref(false)
 
     async function logout() {
-      await userApi.csc.logout()
-      profileMenuOpen.value = false
-      mobileMenuOpen.value = false
-      store.dispatch('setUser', null)
-      router.push({ name: 'Home' })
+      try {
```
paribaker, line 148: This happens every once in a while to me where logout returns a 401 error due to a missing token in the db, then it is impossible to logout

{{cookiecutter.project_slug}}/clients/web/vue3/src/composables/Users.ts
```diff
@@ -0,0 +1,111 @@
...
+    ResetPasswordForm,
+    userApi,
+  } from '@/services/users'
+  import { useMutation, useQueryClient } from '@tanstack/vue-query'
+  import { reactive, ref } from 'vue'
+  import { useRouter } from 'vue-router'
+  import { useStore } from 'vuex'
+  import { useAlert } from '@/composables/CommonAlerts'
+  
+  export function useUsers() {
```
paribaker, line 14: Try using composables for common functionality, this may not be the exact expected implementation but it works nicely. I think ideally, as is done in react, you should have separate _hooks_ for each one.

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.ts
```diff
@@ -1,18 +1,20 @@
...
-  loginShape,
+  userShape,
 } from './models'
 
 const login = createCustomServiceCall({
   inputShape: loginShape,
   outputShape: userShape,
   cb: async ({ client, input, utils }) => {
     const res = await client.post('/login/', utils.toApi(input))
+    console.log(res)
```
oudeismetis, line 17: remove?

{{cookiecutter.project_slug}}/clients/web/vue3/src/services/users/api.ts
```diff
@@ -50,3 +51,14 @@ export const userApi = createApi({
...
+export const organizationApi = createApi({
+  client: axiosInstance,
+  baseUri: '/organizations/',
+  models: {
+    entity: {
+      id: z.string(),
+      name: z.string(),
+    },
+  },
+})
```
oudeismetis, line 64: orgs isn't a thing in the bootstrapper

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/RequestPasswordReset.vue
```diff
@@ -7,51 +7,36 @@
...
-          folder.
-        </p>
-        <div class="pt-6">
-          <button type="button" class="btn--primary bg-primary">
-            <router-link :to="{ name: 'Login' }" class="" id="login-link">
-              Return to Login
-            </router-link>
-          </button>
-        </div>
-      </template>
```
oudeismetis, line 47: So you removed this confirmation logic?

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/ResetPassword.vue
```diff
@@ -24,18 +25,21 @@
           type="password"
           label="Confirm Password"
           placeholder="Confirm Password"
+          :id="form.confirmPassword.id"
+          data-cy=""
```
oudeismetis, line 29: this `data-cy` either needs a value for a test to pass....or it's not needed at all

oudeismetis (approved):

PR #318, opened 2024-06-25
React/RN Clients -  Add `useLogout` hook to wrap logout call in mutation

lakardion:
## What this does

Add a hook for logout to wrap it in mutation and use it safely in components.

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/clients/mobile/react-native/src/stores/navigation.ts
```diff
@@ -0,0 +1,4 @@
+import type { MyNavio } from '@screens/routes'
+import { atom } from 'jotai'
+
```
paribaker, line 3: Where does this get set? It seems to me it is always null

lakardion: True I think I missed that line in the navio file

paribaker:
@lakardion moved this to in progress as it has conflicts

lakardion:
> @lakardion moved this to in progress as it has conflicts

Done @paribaker

paribaker (approved):

PR #320, opened 2024-07-02
RN re-usable button and a preview screen for components

paribaker:
## What this does

It adds a reusable button to the stack that we use regularly. Also adds a preview page to view all available components


{{cookiecutter.project_slug}}/clients/mobile/react-native/src/screens/routes.ts
```diff
@@ -28,10 +29,15 @@ export const tabDefaultOptions = (): BottomTabNavigationOptions => ({
...
 export const navio = Navio.build({
-  screens: { Auth, Login, SignUp, Main, DashboardScreen },
+  screens: { Auth, Login, SignUp, Main, DashboardScreen, ComponentsPreview },
   stacks: {
     AuthStack: ['Auth'],
     MainStack: ['DashboardScreen'],
+    /**
+     * Set me as the root to see the components preview
+     */
+    PreviewStack: ['ComponentsPreview'],
```
oudeismetis, line 39: I think for new projects we should just link to this view from the main screen. It'll then be up to the devs to remove it, hide it, etc.
As it is, I'm concerned it'll be lost/forgotten

oudeismetis (approved):

lakardion (approved):
lgtm

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/screens/routes.ts
paribaker: I was thinking we would just add it to the Main landing page with a new auth screen that we used in other projects, that way its a big clear button that says To the Components Preview!

PR #321, opened 2024-07-04
minimal updates to accommodate django 4.2.x

paribaker:
## What this does

Upgrades Django to Django 4.2.x
Checked python graph to see which packages were breaking, it appears only: 
django-current-user, django-extensions and djangorestframework needed updates

- [x] Django 4.2.x
- [x] django-extensions 3.2.3
- [x] django-current-user 0.7.0
- [x] djangorestframework 3.15.* 

## How to test

### DIFF
--- a/{{cookiecutter.project_slug}}/Pipfile
+++ b/{{cookiecutter.project_slug}}/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-Django = "==4.0.*"
+Django = "==4.2.*"
 django-extensions = "==3.2.3"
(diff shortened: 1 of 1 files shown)


whusterj (commented):
Let's target Django 4.2, since that is the current LTS version: https://www.djangoproject.com/download/

Otherwise looks good. I'm concerned about regressions that may not be obvious or caught by unit tests, but since 3.2 is EOL, I'm inclined to say let's push forward and see what issues come up.

paribaker:
I'll play around with it for a bit, but I did install it to a personal app and it seems to be running fine

whusterj (approved):

PR #328, opened 2024-07-24
Clients - Mobile - Add better rules for RN eslint to prevent raw text in code

lakardion:
## What this does

Add a rule that errors whenever the linter finds strings not wrapped in `<Text>` component which causes crashes on runtime.

Note that here `eslint-plugin-react-native` was already installed but in the project I was working on this wasn't installed and it is a requirement for this lint rule to work:
```
npm install -D eslint-plugin-react-native
```

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


paribaker (approved):
LGTM

PR #335, opened 2024-08-09
[React - Vue] - Migrate E2E Tests from Cypress to Playwright 

noriega2112:
## What this does

Switches the E2E testing libraries from Cypress to Playwright for the React and Vue sides

## Checklist
- [x] Switch to playwright on React Side
- [x] Switch to playwright on Vue Side
- [x] Switch E2E GH workflow

## How to test


{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/sign-up.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const PASSWORD = 'PASSWORD'
+
+test('Login workflow', async ({ page }) => {
+  await page.goto('/sign-up')
+  await page.getByTestId('first-name').fill('playwright')
+  await page.getByTestId('last-name').fill('e2e test')
+  await page
+    .getByTestId('email')
+    .fill(`playwright${Math.floor(Math.random() * 1000)}@thinknimble.com`)
```
whusterj, line 11: I suggest using a UUID or a larger number here to greatly reduce the chance of a collision.

whusterj (commented):
The code looks good, but the Chrome and Firefox e2e tests have not passed yet:

image

Once these are passing and the conflict with `main` is resolved, this should be good to go!

{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/sign-up.spec.ts
noriega2112: yeah you are right, I replaced it for a timestamp instead as its value is unique

noriega2112:
> The code looks good, but the Chrome and Firefox e2e tests have not passed yet:
> 
> image
> Once these are passing and the conflict with `main` is resolved, this should be good to go!

@whusterj  These checks shouldn't be showing as they were part of the `cypress.yml` github action that was before, I think they are still showing as they were required prior to opening the PR. 

Now this is being handled on the playwright config file which makes the tests to run on the 3 major browsers.

whusterj:
> @whusterj These checks shouldn't be showing as they were part of the `cypress.yml` github action that was before, I think they are still showing as they were required prior to opening the PR.
> 
> Now this is being handled on the playwright config file which makes the tests to run on the 3 major browsers.

I see - that makes sense! They should disappear after we merge.

{{cookiecutter.project_slug}}/clients/web/react/tests/e2e/specs/sign-up.spec.ts
whusterj: There is still a chance of collision when browser tests are run in parallel, but that should be very unlikely, so this is a lot better. Thanks!

whusterj (approved):
Looks good to go!

PR #340, opened 2024-08-15
Feature/add expo builds pr

paribaker:
## What this does

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


.github/workflows/mobile.yml
```diff
@@ -86,11 +86,42 @@ jobs:
...
\ No newline at end of file
+          EXPO_PUBLIC_ROLLBAR_ACCESS_TOKEN: "1a19e5da05b2435b802d5a81aba2bbd7"
+      
+      - name: Update EAS Config
+        working-directory: ./my_project/mobile
+        run: |
+            jq --arg pr_number "$PR_NUMBER" --arg expo_channel "$EXPO_CHANNEL" '.build["review_" + $pr_number] = {
+              "developmentClient": true,
+              "distribution": "internal",
+              "channel": $expo_channel,
```
paribaker, line 110: One thought I had was, if we remove the channel here will we be able to use the statically defined config instead of dynamically creating a new entry?
The reason we are adding this entry is so that there will be a unique build for this PR that does not affect other PR's. However since this one is built on the fly in order to test it the user will have to: 
Register their device, and update the cert to and then rebuild. 

1. eas device:create 
2. eas credentials ---> review_pr-#
3. re build

PR #341, opened 2024-08-15
Clients - Add default parameter serializer for arrays on axios.

### LINKED ISSUE #334: Clients - Add default array parameter serializer to axios
Prevent axios to do the default param serialization of array parameters which are not what we usually expect in our django backends.

Use `qs` to get the either a comma separated parameter or a parameter repetition pattern.

lakardion:
## What this does

Closes #334 

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/clients/web/react/src/services/axios-instance.ts
```diff
@@ -9,6 +10,9 @@ export const axiosInstance = axios.create({
   headers: {
     'Content-Type': 'application/json',
   },
+  paramsSerializer: (params) => {
+    return qs.stringify(params, { arrayFormat: 'repeat' })
```
paribaker, line 14: hmm I think we had said the backend prefers csv, this is set to repeat?

lakardion: This was the only example I had. We can set it to repeat instead if that's the case

paribaker: I think it would be better but the test I just run real quick seems to indicate that it works either way! (but that was a quick test)

lakardion: Maybe django understands both. Just updated for it to be comma

paribaker (approved):

PR #342, opened 2024-08-15
React Native - Add fonts to `nativewind` and remove custom `Text` component

### LINKED ISSUE #327: Clients - React Native - Remove custom `Text`component in favor of defining custom NativeWind fonts
There seems to be a way to define custom fonts class names with nativewind so that we avoid having to do the mess that it is right now `fonts.ts` and needing the custom `Text` component.

https://github.com/nativewind/nativewind/issues/964#issuecomment-2224953451

https://github.com/nativewind/nativewind/issues/923#issuecomment-2136665869

We could probably also use the `useFonts` hook instead of the `loadAsync`. However I'm not exactly sure whether that would be an improvement

lakardion:
## What this does

Remove custom Text component which reason for existing was using custom fonts. Now fonts are loaded into NativeWind and we can use the built-in `Text` component from `react-native`

Closes #327 

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/clients/mobile/react-native/tailwind.config.js
```diff
@@ -7,6 +7,18 @@ module.exports = {
...
+        'primary-black-italic': `Montserrat-BlackItalic`,
+        'primary-bold': `Montserrat-Bold`,
+        'primary-bold-italic': `Montserrat-BoldItalic`,
+        'primary-italic': `Montserrat-Italic`,
+        'primary-light': `Montserrat-Light`,
+        'primary-light-italic': `Montserrat-LightItalic`,
+        'primary-medium': `Montserrat-Medium`,
+        'primary-medium-italic': `Montserrat-MediumItalic`,
+        'primary-regular': `Montserrat-Regular`,
+      },
```
lakardion, line 21: Now for referencing any of this we just need to do `font-primary-<variant>`

paribaker (commented):
so this is doing two things 1. switching to expo fonts AND moving them to the tw config. I assume that means in order for the second to work the first had to be done as well?

lakardion:
> so this is doing two things 1. switching to expo fonts AND moving them to the tw config. I assume that means in order for the second to work the first had to be done as well?

No. We already were using expo fonts. So that's the same. 

We're just remapping how we use the different font variants. Rather than doing it with a custom component we're embedding them into native wind class names

paribaker (approved):

PR #343, opened 2024-08-20
Add Code Password Reset to the Bootstrapper

eliasbiagioninc:
## What this does

Switch from password reset email links to 7-digit codes for resetting passwords. Discussion [Add Code Password Reset to the Bootstrapper](https://github.com/thinknimble/tn-spa-bootstrapper/discussions/339)

## Checklist
- [x] Remove token management for reset password
- [x] Add models to store OTP codes
- [x] Update reset password endpoint to generate new codes
- [x] Update reset password confirmation to allow code, password and email
- [x] Add screens - Mobile
- [x] Add screens - Web
  - [x] React
  - [x] Vue
- [x] Tests

## How to test


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
```diff
@@ -89,3 +93,48 @@ def reset_password_context(self):
...
+    code = models.CharField(max_length=255)
+    is_used = models.BooleanField(default=False)
+    objects = UserResetPasswordCodeMessagesManager()
+
+    def __str__(self):
+        return f"reset password code for {self.user.email}"
+
+    @property
+    def is_valid(self):
+        return not (self.is_used | (self.created > now() + timedelta(minutes=settings.RESET_PASSWORD_CODE_VALIDITY_MINUTES)))
```
oudeismetis, line 137: if a code is used, wouldn't we just delete the model instance?

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/serializers.py
```diff
@@ -69,3 +75,24 @@ def validate(self, data):
...
+class ResetPasswordSerializer(serializers.Serializer):
+    code = serializers.CharField(allow_blank=False, required=True)
+    password = serializers.CharField(allow_blank=False, required=True)
+
+    def validate_code(self, value):
+        code_from_db = (
+            self.context.get("user")
+            .reset_password_codes.filter(created__gte=(timezone.now() - timedelta(minutes=settings.RESET_PASSWORD_CODE_VALIDITY_MINUTES)))
+            .first()
+        )
```
oudeismetis, line 89: shouldn't they only ever have a single code at any moment? If it expires, then it fails validation.
If they request a new one, the old one should be deleted.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/signals.py
```diff
@@ -1,11 +1,36 @@
...
 from django.dispatch import receiver
 from rest_framework.authtoken.models import Token
 
 from {{ cookiecutter.project_slug }}.core.models import User
+from {{ cookiecutter.project_slug }}.utils.emails import send_html_email
+
+from .dispatchers import new_reset_password_code_created_ds
+
+# Logger
+logger = logging.getLogger(__name__)
```
oudeismetis, line 14: we know it's a logger. No need for a code comment

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -119,14 +119,12 @@ def test_password_reset(caplog, api_client, sample_user):
...
     rf = RequestFactory()
     post_request = rf.post("api/password/reset/", {"email": sample_user.email})
-    request_reset_link(post_request)
-
-    # Grab from the logs the actual URL link we would send to the user
-    password_reset_creds = caplog.text.split("password/reset/confirm/")[1].split('"')[0]
-    password_reset_url = f"/api/password/reset/confirm/{password_reset_creds}/"
+    request_reset_code(post_request)
 
+    code = caplog.text.split()[-1].replace('"', "")
```
oudeismetis, line 124: were getting the code from the caplog??

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -92,43 +94,38 @@ def destroy(self, request, *args, **kwargs):
...
-    if not user or not token:
-        raise ValidationError(detail={"non-field-error": "Invalid or expired token"})
-    is_valid = default_token_generator.check_token(user, token)
-    if not is_valid:
-        raise ValidationError(detail={"non-field-error": "Invalid or expired token"})
-    logger.info(f"Resetting password for user {user_id}")
+    email = kwargs.get("email")
+    user = User.objects.filter(email=email).first()
+    if not user:
+        raise ValidationError(detail={"non_field_errors": ["User not found with that email"]})
```
oudeismetis, line 116: Security issue. Someone could use this to determine what users are in the DB.

I would just fail validation with a generic error. The error should be the same regardless of the reason (ex: bad code vs bad email)

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/migrations/0002_userresetpasswordcodemessages.py
lakardion, line 1: For my local project. This migration was not enough ( I tried sending the email and it responded with 500)
I had an existing database, but I think there's a missing migration.
And when I ran the server, it said that there were unapplied changes to the models. So I tried `makemigrations`

```
It is impossible to add a non-nullable field 'code' to userresetpasswordcodemessages without specifying a default. This is because the database needs something to populate existing rows.
Please select a fix:
 1) Provide a one-off default now (will be set on all existing rows with a null value for this column)
 2) Quit and manually define a default value in models.py.
Select an option: 2
```

PR #347, opened 2024-08-28
React Native - Add Settings Stack

noriega2112:
## What this does

## Checklist
- [x] add main settings screen
- [x] add contact screen
- [x] add edit profile screen
- [x] enable user deletion


{{cookiecutter.project_slug}}/clients/mobile/react-native/src/screens/settings/edit-profile.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const parsedName = fullNameZod.safeParse(fullName)
+  const isValid = parsedName.success
+  const qClient = useQueryClient()
+
+  const { mutate: save, isPending: isSaving } = useMutation({
+    mutationFn: userApi.csc.patchUser,
+    onMutate: async () => {
+      const userSnapshot = qClient.getQueryData<UserShape>(userQueries.all())?.fullName
+      await qClient.cancelQueries({ queryKey: userQueries.all() })
+      await qClient.setQueryData(userQueries.all(), (input?: UserShape) => {
```
lakardion, line 46: I think this should be userQueries.retrieve(userId) ?

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/screens/settings/edit-profile.tsx
```diff
@@ -0,0 +1,201 @@
...
+        return input ? { ...input } : undefined
+      })
+      return { userSnapshot }
+    },
+    onSuccess: () => {
+      qClient.invalidateQueries({ queryKey: userQueries.all() })
+    },
+    onError: (e, _, context) => {
+      //rollback update
+      qClient.setQueryData(userQueries.all(), (input?: UserShape) => {
```
lakardion, line 56: same, I think the ['users',userId] is where the query value is stored

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
```diff
@@ -34,12 +34,22 @@ const logout = createCustomServiceCall(async ({ client }) => {
...
+const patchUser = createCustomServiceCall({
+  inputShape: partialUserShape,
+  outputShape: userShape,
+  cb: async ({ client, slashEndingBaseUri, input, utils }) => {
+    const { id, ...body } = utils.toApi(input)
+    const res = await client.put(`${slashEndingBaseUri}${id}/`, body)
+    return utils.fromApi(res.data)
+  },
+})
+
```
lakardion, line 46: Why not the built-in `update` ?
It is a patch and I think it does exactly this what you're doing in the custom call

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/hooks.ts
```diff
@@ -6,28 +6,21 @@
 import { useEffect } from 'react'
 import { useQuery, useMutation } from '@tanstack/react-query'
 import { useAuth } from '@stores/auth'
-import { userApi } from './api'
+import { userApi, userQueries } from '.'
```
lakardion, line 9: do not import from `.` it sometimes fails to resolve depending on the importing order and module resolution

Even though this is done by VS code automatically check that when you import from the same folder reference the file and not the index

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/api.ts
lakardion: also remove the partial user shape

{{cookiecutter.project_slug}}/clients/mobile/react-native/src/services/user/models.ts
```diff
@@ -38,3 +39,24 @@ export const loginShape = {
...
+}
+
+export const fullNameZod = z.string().refine(
+  (value) => {
+    return value.split(' ').filter(Boolean).length >= 2
+  },
+  { message: 'Please provide a full name (first and last name)' },
+)
+
+export const partialUserShape = { ...zodPartialize(userShape), id: z.string().uuid() }
```
lakardion, line 62: this could potentially go all away if we just use the `update`

lakardion (dismissed):

whusterj (approved):
All checks are passing now! I think this is good to go if @lakardion signs off.

whusterj:
These checks are still showing up, and I'm not sure why:

image

lakardion:
The app looks good with the fonts so this should be good to go

noriega2112:
> These checks are still showing up, and I'm not sure why:
> 
> image

@whusterj  could it be a required check configured on the repo itself?

lakardion (approved):

lakardion:
> These checks are still showing up, and I'm not sure why:
> 
> image

Just made them to be gone. They were being required from Settings in the branch protection rules. Added playwright instead

> @lakardion - Did you want to see changes made before merging, or can I go ahead and merge this in?

Ah yeah I approved since the comments were that big but Edwin fixed them I just reapproved! This should be good to go

whusterj:
Awesome - thanks to you both!

PR #348, opened 2024-08-29
Improve documentation for Playwright tests.

whusterj:
## What this does

- Changes some missed references to `CYPRESS` to `PLAYWRIGHT` in `.env.example`, and corrects the env variable name for `PLAYWRIGHT_TEST_BASE_URL`.
- Adds additional installation steps to README.md that are necessary to run Playwright locally


whusterj (commented):
I think that's the last of the references to Cypress. Also added missing install steps for Playwright.

PR #246, opened 2023-10-24
Expo Build for PRs

oudeismetis:
## What this does

Sometimes you need/want to do a full build test for a PR. Creating a Github Action to do that when manually built.

## Checklist
- [ ] Initial version guessing at implementation
- [ ] Add support for choosing IOS and/or Android
- [ ] Test

## How to test

Add user steps to achieve desired functionality for this feature.

PR #349, opened 2024-08-30
Mobile - Add expo workflow yaml for allowing devs to trigger manual builds

lakardion:
## What this does

With this any dev in the repo can create builds following any given profile from `eas.json`

This allows to easily create development builds or staging builds without the need of the dev to have to go through eas CLI in the local.

## How to test


{{cookiecutter.project_slug}}/.github/workflows/expo-build.yml
```diff
@@ -0,0 +1,66 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      profile:
+        description: "Build profile"
+        required: true
+        default: "development"
+        type: choice
+        # Should match the ones in `eas.json`
+        options:
```
paribaker, line 13: this goes nicely with my open PR as well that creates dynamic review app based builds. 
I think dynamically adding them to the eas.json like this allows us to create unique builds. In this case as you have it, each engineer's build will overwrite the others (although at this moment many of us don't really need builds that often). 
I also wonder if this can be avoided by removing the channel arg

lakardion: I mean initially I wrote this because I happened not to have enough permissions in a project to use the eas cli myself so I just used the action to get a build  I needed.
This as you said is not something that we would use often, it is just a convenience, mostly to get development builds that we usually need when we deviate from expo go.

{{cookiecutter.project_slug}}/.github/workflows/expo-build.yml
```diff
@@ -0,0 +1,66 @@
...
+          node-version: 18.x
+          cache: npm
+          cache-dependency-path: ./mobile/package-lock.json
+
+      - name: 🏗 Setup Expo and EAS
+        uses: expo/expo-github-action@v8
+        with:
+          expo-version: latest
+          eas-version: latest
+          token: ${{ secrets.EXPO_TOKEN }}
```
whusterj, line 53: @lakardion - Cookiecutter is trying to render this and failing to 'bake'.

lakardion: Whoops, let's see now

whusterj:
This duplicates #246 — I think it's a good idea along with Pari's idea for dynamic builds for PRs in #340. I'm going to close this PR and move this work over into #246.

whusterj (approved):
LGTM!

PR #350, opened 2024-09-05
Remove jinja2-time from _extensions

whusterj:
## What this does

According to https://github.com/cookiecutter/cookiecutter/issues/1888, the jinja2-time extension is no longer required and can be removed. In more recent versions of cookiecutter, the time extension is loaded automatically.

This should make it possible to run the bootstrapper with the following commands. `pipx` is recommended by cookiecutter's docs:

```bash
brew install pipx
pipx install cookiecutter
pipx run cookiecutter gh:thinknimble/tn-spa-bootstrapper
```

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,7 +18,6 @@ pytest-instafail = "==0.4.2"
 PyYAML = "==6.0.1"
 PyGithub = "==1.55"
 Jinja2 = "==3.0.1"
-jinja2-time = "*"
 
 [dev-packages]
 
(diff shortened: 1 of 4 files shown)


whusterj:
Tests are passing and it appears to be working locally.

Tested with:

```
brew install pipx

pipx install cookiecutter
pipx run cookiecutter https://github.com/thinknimble/tn-spa-bootstrapper.git --checkout remove-jinja2-time
```

PR #352, opened 2024-09-09
Backend Security Audit fixes

oudeismetis:
## What this does

Had to do a security audit recently for one of our projects that uses this bootstrapper.
This PR includes fixes from that effort.

## How to test

### DIFF
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/models.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/models.py
@@ -16,4 +16,4 @@ class AbstractBaseModel(models.Model):
         abstract = True
 
     def __str__(self):
-        return "ah yes"
+        return "__str__ not defined for this model"
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
@@ -35,7 +35,15 @@ class CustomUserAdmin(UserAdmin):
             },
         ),
     )
-    list_display = ("email", "permissions", "is_active", "is_staff", "is_superuser", "first_name", "last_name",)
+    list_display = (
+        "email",
+        "permissions",
+        "is_active",
+        "is_staff",
+        "is_superuser",
+        "first_name",
+        "last_name",
+    )
     list_display_links = (
         "is_active",
         "email",
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
@@ -24,7 +24,7 @@ class CustomUserAdmin(UserAdmin):
                 )
             },
         ),
-        ("Admin Options", {"classes": ("collapse",), "fields": ("is_staff", "groups")}),
+        ("Admin Options", {"classes": ("collapse",), "fields": ("is_active", "is_staff", "is_superuser", "groups")}),
     )
     add_fieldsets = (
         (
@@ -35,7 +35,7 @@ class CustomUserAdmin(UserAdmin):
             },
         ),
     )
-    list_display = ("email", "first_name", "last_name", "is_active", "is_staff", "is_superuser", "permissions")
+    list_display = ("email", "permissions", "is_active", "is_staff", "is_superuser", "first_name", "last_name",)
     list_display_links = (
         "is_active",
         "email",
@@ -57,7 +57,7 @@ class CustomUserAdmin(UserAdmin):
     ordering = []
 
     def permissions(self, obj):
-        return ", ".join([g.name for g in obj.groups.all()])
+        return [g.name for g in obj.groups.all()]
 
     class Media(AutocompleteAdminMedia):
         pass
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
@@ -11,10 +11,25 @@ from {{ cookiecutter.project_slug }}.utils.sites import get_site_url
 logger = logging.getLogger(__name__)
 
 
+class UserQuerySet(models.QuerySet):
+    def for_user(self, user):
+        if not user or user.is_anonymous:
+            return self.none()
+        elif user.is_staff:
+            return self.all()
+        return self.filter(pk=user.pk)
+
+
 class UserManager(BaseUserManager):
     """Custom User model manager, eliminating the 'username' field."""
 
     use_in_migrations = True
+
+    def get_queryset(self):
+        return UserQuerySet(self.model, using=self.db)
+
+    def for_user(self, user):
+        return self.get_queryset().for_user(user)
 
     def _create_user(self, email, password, **extra_fields):
         """
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/permissions.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/permissions.py
@@ -1,8 +1,8 @@
 from rest_framework import permissions
 
 
-class CreateOnlyPermissions(permissions.BasePermission):
-    def has_permission(self, request, view):
-        if view.action == "create":
-            return True
-        return False
+class HasUserPermissions(permissions.BasePermission):
+    """Admins should be able to perform any action, regular users should be able to edit and delete self."""
+
+    def has_object_permission(self, request, view, obj):
+        return request.user.is_authenticated and (request.user.is_staff or obj == request.user)
--- a/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
+++ b/{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
@@ -35,6 +35,13 @@ def test_create_user():
 
 
 @pytest.mark.django_db
+def test_create_user_api(api_client):
+    data = {"email": "example@example.com", "password": "password", "first_name": "Test", "last_name": "User"}
+    res = api_client.post("/api/users/", data, format="json")
+    assert res.status_code == status.HTTP_201_CREATED, res.data
+
+
+@pytest.mark.django_db
 def test_create_superuser():
     superuser = User.objects.create_superuser(email="test@example.com", password="password", first_name="Leslie", last_name="Burke")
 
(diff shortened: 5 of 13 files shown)


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -50,13 +50,7 @@ def post(self, request, *args, **kwargs):
...
 
 
-class UserViewSet(
-    viewsets.GenericViewSet,
-    mixins.RetrieveModelMixin,
-    mixins.ListModelMixin,
-    mixins.UpdateModelMixin,
-    mixins.DestroyModelMixin,
-):
+class UserViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin):
```
oudeismetis, line 53: We implement some of these with overrides below.
Oddly, by implementing `update()` and having the `UpdateModelMixin` above, we ended up with PATCH and POST endpoints that both handled updating, but had different rules and validation.
Most projects shouldn't need all of these (ex: listing all users). If they do, it's easy enough to add them, along with the proper security controls.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -66,29 +60,29 @@ class UserViewSet(
...
+    def destroy(self, request, *args, **kwargs):
+        """
+        When deleting a user's account, just disable their account first
+        The user may have a regret and try to get their account back
+        A background job should then properly delete the data after X days
+        """
+        user = self.get_object()
+        user.is_active = False
+        user.save()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
oudeismetis, line 85: Comment above should be clear.
By default apps shouldn't just delete important data from the DB like this.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -390,11 +392,6 @@ def filter(self, record):
 # Popular testing framework that allows logging to stdout while running unit tests
 TEST_RUNNER = "django_nose.NoseTestSuiteRunner"
 
-CORS_ALLOWED_ORIGINS = ["https://{{ cookiecutter.project_slug|replace('_', '-') }}-staging.herokuapp.com", "https://{{ cookiecutter.project_slug|replace('_', '-') }}.herokuapp.com"]
-{% if cookiecutter.client_app.lower() != 'none' -%}
-CORS_ALLOWED_ORIGINS.append("http://localhost:8080")
-{% endif -%}
-
```
oudeismetis, line 397: See further up.
This is now controlled by `ALLOWED_HOSTS`. Hardcoding values like localhost into this file is a security risk.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/urls.py
```diff
@@ -1,6 +1,9 @@
 from django.contrib import admin
 from django.urls import include, path
 
+admin.site.site_header = "{{ cookiecutter.project_name }} Admin"
+admin.site.site_title = "{{ cookiecutter.project_name }}"
+
```
oudeismetis, line 6: Just a nice little addition so that managing multiple projects makes it more obvious which admin you are in.

Also, non-dev project stakeholders don't know (or need to care) what the "Django Admin" is. It's just the admin for the app.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/models.py
```diff
@@ -16,4 +16,4 @@ class Meta:
         abstract = True
 
     def __str__(self):
-        return "ah yes"
+        return "__str__ not defined for this model"
```
oudeismetis, line 19: gentle nudge for devs to set this for their new models

whusterj: Thank you

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -50,45 +50,39 @@ def post(self, request, *args, **kwargs):
...
-    mixins.DestroyModelMixin,
-):
+class UserViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin):
     queryset = User.objects.all()
     serializer_class = UserSerializer
 
     # No auth required to create user
     # Auth required for all other actions
-    permission_classes = (permissions.IsAuthenticated | CreateOnlyPermissions,)
+    permission_classes = (HasUserPermissions,)
```
oudeismetis, line 59: `IsAuthenticated` was causing a problem here. It meant that a logged in user could do a GET on any user.
Changing this to a permission class that only lets them take action on themselves by default (unless they are a staff user)

Create still works fine.

There are automated tests now for all of this so should this break on a project, a dev would have to also remove the test to get around it.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -50,45 +50,45 @@ def post(self, request, *args, **kwargs):
...
 
-    @transaction.atomic
-    def create(self, request, *args, **kwargs):
+    def get_queryset(self):
         """
-        Endpoint to create/register a new user.
+        Users should only find themselves by default
         """
+        user = self.request.user
+        return User.objects.filter(pk=user.pk)
```
oudeismetis, line 65: @whusterj I removed the querset and replaced it with this call. It might not matter since we don't have list as an option here anymore. But probably a good default to have for projects that try adding a quick mixin here.

oudeismetis: aaaand it's broken the tests. Will investigate in the morning.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/tests.py
```diff
@@ -50,7 +58,59 @@ def test_create_user_from_factory(sample_user):
...
+    assert res.data["email"] == sample_user.email
+
+
+@pytest.mark.django_db
+def test_get_other_user(api_client, sample_user, user_factory):
+    api_client.force_authenticate(sample_user)
+    other_user = user_factory()
+    other_user.save()
+    res = api_client.get(f"/api/users/{other_user.pk}/")
+    assert res.status_code == status.HTTP_404_NOT_FOUND
```
oudeismetis, line 90: Note that this is a 404. The other user does exist, but we don't want to make bad actors aware of that fact. So it's a 404.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -50,45 +50,45 @@ def post(self, request, *args, **kwargs):
...
+    permission_classes = (HasUserPermissions,)
 
-    @transaction.atomic
-    def create(self, request, *args, **kwargs):
+    def get_queryset(self):
         """
-        Endpoint to create/register a new user.
+        Users should only find themselves by default
         """
+        return super().get_queryset().for_user(self.request.user)
```
oudeismetis, line 65: This `get_queryset` doesn't do anything right now. But if you were to add the list mixin or add a `def list()` below then it'll do the job of only returning that logged in user for that list (or all users in a list for staff users)

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
oudeismetis: I had to add the queryset back. Left a comment as to why

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
```diff
@@ -50,45 +50,45 @@ def post(self, request, *args, **kwargs):
...
 
-class UserViewSet(
-    viewsets.GenericViewSet,
-    mixins.RetrieveModelMixin,
-    mixins.ListModelMixin,
-    mixins.UpdateModelMixin,
-    mixins.DestroyModelMixin,
-):
+class UserViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin):
     queryset = User.objects.all()
```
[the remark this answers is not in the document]
whusterj: Ah, in that case I think you could replace it with `User.objects.none()`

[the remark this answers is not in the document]
oudeismetis: I like it, but that doesn't work. A couple of tests are failing now because the chain of filters starts by finding nothing.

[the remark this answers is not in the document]
oudeismetis: setting it to just `User.objects` works though. And that would imply to devs that filters get layered on elsewhere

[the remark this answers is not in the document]
whusterj: Yeah, I think that's a good place to land

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
```diff
@@ -57,7 +65,7 @@ class CustomUserAdmin(UserAdmin):
     ordering = []
 
     def permissions(self, obj):
-        return ", ".join([g.name for g in obj.groups.all()])
+        return [g.name for g in obj.groups.all()]
```
whusterj, line 68: @oudeismetis - I trust that you checked that this works? This is changing the return type from `str` to `list[str]`.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/models.py
```diff
@@ -11,11 +11,26 @@
 logger = logging.getLogger(__name__)
 
 
+class UserQuerySet(models.QuerySet):
+    def for_user(self, user):
+        if not user or user.is_anonymous:
+            return self.none()
+        elif user.is_staff:
+            return self.all()
+        return self.filter(pk=user.pk)
```
whusterj, line 20: 👍🏻

whusterj (approved):
Just one question on final review. If the answer is "yes," then this is good to go.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/admin.py
oudeismetis: correct. I pulled this from a project where I made the change.
The join isn't needed and happens for you

oudeismetis: Just re-tested on a locally generated my_project.
image

whusterj: Great! This is good to go then.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/permissions.py
```diff
@@ -1,8 +1,8 @@
 from rest_framework import permissions
 
 
-class CreateOnlyPermissions(permissions.BasePermission):
-    def has_permission(self, request, view):
-        if view.action == "create":
-            return True
-        return False
+class HasUserPermissions(permissions.BasePermission):
```
paribaker, line 4: Q, is there a situation where `request.user` might be None? if so wont that cause a 500 error crash

oudeismetis: Should not happen ever

oudeismetis: Notice that the viewset where this is used also handles user creation requests. There are automated tests that cover that case.

I do think I've had to handle some sort of none case in the past, but I haven't seen that in a long while. But should that bug arise it would be a small/quick/obvious fix

whusterj: I think the Authentication Middleware always adds `user` to the request. If the client is unauthenticated, then it sets `request.user` to an instance of `AnonymousUser`. But as Ed said, it should never be `None` or falsey. Could only happen if that middleware is removed.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/views.py
paribaker: Are you sure? I think get_queryset is also called for update methods

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/core/permissions.py
oudeismetis: That tracks. Otherwise `is_anonymous` wouldn't be very useful. That might have been a bug in older versions that people reported

paribaker: ah nvm this is used on the object level!

PR #354, opened 2024-09-13
Decouple session management of tokens from user data

oudeismetis:
## What this does

Our current implementation is fine, but becomes a problem the moment you want to update the users stored info without also updating their token.

### Current client statuses
Main problem is Vue.

Both React and React Native, don't depend on the user model for getting the token. Upon authentication we set the token in store as well as the userId.
Therefore the userId is used to fetch the user information with `users/{userid}`
However, since the user is returned by the auth call, we save that information in the query cache to avoid needing to call this endpoint, but the source of truth for the user information is always the server and not the store.

## Checklist
- [x] Separate Vue calls for user data and user token
- [ ] Refactor to have only 1 endpoint/model/shape have a token in it
- [ ] Update Vue
- [ ] Update React
- [ ] Update React Native

## How to test


whusterj:
@paribaker and @lakardion - You might be best equipped to help Ed with this one.

{{cookiecutter.project_slug}}/clients/web/vue3/src/stores/auth.ts
lakardion, line 1: moved store to setup store which very much looks like a composobale

lakardion:
@oudeismetis I removed the coupling in Vue which was the only client that was having both user and token combined in the same state.
After the new endpoint for the token is ready we can update all the clients to use it.

So right now the two places we're getting a token for the app are login and sign-up. 
For login we should just send the user and pass to get the token. 
Then for sign-up we should do the post for creating the user and subsequently perform a login (aka get the token with the user and pass) ?

paribaker:
Closing because we removed support for Vue

whusterj:
> Closing because we removed support for Vue

Thanks!

PR #355, opened 2024-09-23
Prevent bad actors injecting behavior into emails

oudeismetis:
## What this does

See title.
The PR is to implement a couple of safeguards to avoid this kind of issue. 

## Checklist
- [ ] Confirm if Django auto-escapes by default
- [ ] Try to reproduce the attack I saw recently
- [ ] Investigate validating all form string inputs for common escape code patterns

## How to test


oudeismetis:
Tested and confirmed that this isn't relevant

PR #361, opened 2024-11-08
Clients - Vue Updates

lakardion:
## Checklist
- [x] Move to SFCs (more concise and the current tendency)
- [x] Use Typescript for real in vue components
- [x] Move any JS to TS
- [x] Remove class-based styling, instead componentize and use tailwind
- ~Erradicate use of margins. Use paddings instead.~ I think I'm gonna leave this for another PR

### DIFF
--- a/{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Home.vue
+++ b/{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Home.vue
@@ -2,9 +2,11 @@
 
 <template>
   <div class="mx-auto flex min-h-full max-w-2xl flex-col items-center px-6 py-20 sm:py-48 lg:px-8">
-    <h1 class="text-4xl font-bold tracking-tight text-gray-900 sm:text-6xl">Welcome to vuets!</h1>
+    <h1 class="text-4xl font-bold tracking-tight text-gray-900 sm:text-6xl">
+      Welcome to {{ cookiecutter.project_name }}!
+    </h1>
     <p class="mt-6 text-lg leading-8 text-gray-600">
-      Here's some information about vuets. Please update and expand on this text. This text is the
+      Here's some information about {{ cookiecutter.project_name }}. Please update and expand on this text. This text is
       first thing that users will see on the home page.
     </p>
     <div class="mt-10 flex items-center justify-center gap-x-6">


oudeismetis:
Good call on skipping the margin issue for now, this PR is already big enough
Will wait on Maya's opinion, but I'm taking a look now. Thank you for doing this 🙏

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Signup.vue
```diff
@@ -1,3 +1,22 @@
...
+const { register, loading, registerForm: form } = useUsers()
+const onRegister = () => {
+  const createAccountInput = {
+    firstName: form.firstName.value ?? '',
+    lastName: form.lastName.value ?? '',
+    email: form.email.value ?? '',
+    password: form.password.value ?? '',
+    confirmPassword: form.confirmPassword.value ?? '',
+  }
+  register(createAccountInput)
```
oudeismetis, line 16: Shouldn't form validation handle this and we should just be able to pass the form to the API call? Why do we need to manually extract each field?

lakardion: This is a known bad interaction between tn-forms and typescript.
Even after validating the type of the form value is `<type> | undefined`
Ideally, after form validation,  the `form.value` should have the right required field types
I'm [pushing for a good TS update](https://github.com/thinknimble/tn-forms/pull/26) to tn-forms but I don't think it is something that will come soon 😅

{{cookiecutter.project_slug}}/clients/web/vue3/src/components/Button.vue
```diff
@@ -0,0 +1,46 @@
...
+  }
+> = {
+  primary: { buttonClass: 'bg-primary text-white', spinnerColor: 'white' },
+  secondary: { buttonClass: 'bg-secondary text-black hover:bg-primary hover:text-white' },
+  unstyled: {
+    buttonClass: '',
+  },
+}
+
+const props = withDefaults(defineProps<ButtonProps>(), {
```
paribaker, line 27: neat I was getting annoyed with the no defaults thing!

{{cookiecutter.project_slug}}/clients/web/vue3/src/composables/use-users.ts
```diff
@@ -14,15 +18,23 @@ import { reactive, ref } from 'vue'
...
+import { MustMatchValidator } from '@thinknimble/tn-forms'
 
 export function useUsers() {
   const userStore = useUserStore()
   const router = useRouter()
   const qc = useQueryClient()
-  const loginForm = reactive(new LoginForm({}))
-  const forgotPasswordForm = reactive(new EmailForgotPasswordForm({}))
-  const resetPasswordForm = reactive(new ResetPasswordForm({}))
-  const registerForm = reactive(new AccountForm({}))
```
paribaker, line 25: ahh need to fix this, we dont need to use the FormLevelValidator here and can just add it as it was

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Home.vue
```diff
@@ -1,22 +1,18 @@
+<script lang="ts" setup></script>
+
 <template>
   <div class="mx-auto flex min-h-full max-w-2xl flex-col items-center px-6 py-20 sm:py-48 lg:px-8">
-    <h1 class="text-4xl font-bold tracking-tight text-gray-900 sm:text-6xl">
-      Welcome to {{ cookiecutter.project_name }}!
-    </h1>
+    <h1 class="text-4xl font-bold tracking-tight text-gray-900 sm:text-6xl">Welcome to vuets!</h1>
     <p class="mt-6 text-lg leading-8 text-gray-600">
```
paribaker, line 6: I think you accidentally replaced the cookiecutter bindings with your project

{{cookiecutter.project_slug}}/clients/web/vue3/src/views/Signup.vue
paribaker: Yea the issue is that the actual definition of the form field values in the form is 
`<type>|undefined`
Because value is an optional field 
In addition to Dami's work I was thinking we could just add a type caster to avoid the issue

paribaker (dismissed):
Some of the bindings changed, can we also add the updated tn-models here with the .readonly

lakardion:
For tn models upgrade check #369

mayakeeley (approved):

Issue #362, opened 2024-11-09
Upgrade Python


whusterj:
Upgrade to at least Python 3.12, [which is supported through 2028](https://devguide.python.org/versions/).

I'm testing this upgrade on a new project and will bring those changes into a bootstrapper PR.

## Known Compatibility Issues

- Django-nose not supported, maintenance is iffy: https://github.com/jazzband/django-nose/issues/331

PR #363, opened 2024-11-09
Upgrade to Python 3.12

whusterj:
## What this does

This PR upgrades the bootstrapper itself and the resulting generated project from Python 3.10 to Python 3.12. 

Related Issue: #362

## Checklist

- [x] Change all Python versions from 3.10 to 3.12
- [x] Remove Django-nose due to incompatibility
- [x] Identify and fix other problems


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/tests.py
```diff
@@ -37,4 +37,4 @@ def test_get_site_url_negative(settings, custom_settings):
 def test_password_reset_email_link(user):
     context = user.reset_password_context()
     html_body = get_html_body("registration/password_reset.html", context)
-    assert f"{ context['site_url'] }/password/reset/confirm/{ context['user'].id }/{ context['token'] }" in html_body
+    assert f"{context['site_url']}/password/reset/confirm/{context['user'].id}/{context['token']}" in html_body
```
whusterj, line 40: Not sure how this wasn't failing linting before — possibly due to the flake8 upgrade.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
```diff
@@ -389,9 +388,6 @@ def filter(self, record):
     LOGGING["loggers"]["django"]["handlers"].append("rollbar")
     LOGGING["loggers"]["{{ cookiecutter.project_slug }}"]["handlers"].append("rollbar")
 
-# Popular testing framework that allows logging to stdout while running unit tests
-TEST_RUNNER = "django_nose.NoseTestSuiteRunner"
```
whusterj, line 393: This is the most impactful change. Django-nose is not compatible with 3.12, and its maintenance status is questionable, so I think it's best to remove it. I don't think it will make a huge difference to our dev and CI workflows, but let me know if I'm missing something.

oudeismetis: thank you for starting this PR! 🔥

oudeismetis (approved):

oudeismetis (approved):
Looks great!
Always tempted to go right to the newest version (3.13), but I think we should always be ~1 version behind. I just watched a cool demo of a new feature in 3.13, but the video had to call out that their type checker was failing because the feature was so new that the library didn't know it existed.

So probably version bump python once every ~1 year(?) to a non-newest version.

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/settings.py
oudeismetis: testing this on a project now. So far this seems to be a change that doesn't matter

{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/utils/tests.py
oudeismetis: I think the issue here is that flake8 is pinned at the bootstrapper level Pipfile, but for generated projects it's on `*`. The upgrade of python must have triggered a newer version of flake8 that we didn't have access to with python 3.10.

PR #364, opened 2024-11-09
Fix Playwright GH Action

whusterj:
## What this does

Playwright E2E tests are failing in CI on my newly-bootstrapped project with the error below.

I don't think the error is important. I believe the root cause is a working directory issue. This updates the GH Action config with working directory changes that have worked on other projects.

The error:

```bash
SyntaxError: /Users/william/thinknimble/darksaber/client/src/components/HelloWorld.vue: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>? (79:0)

  77 | </template>
  78 |
> 79 | <script>
     | ^
  80 | export default {
  81 |   name: 'HelloWorld',
  82 |   props: {

   at client/tests/unit/example.test.ts:79
```

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/playwright.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/playwright.yml
@@ -21,8 +21,8 @@ jobs:
         working-directory: ./client
         run: npx playwright test --reporter=html
         env:
-          PLAYWRIGHT_TEST_BASE_URL: {{ ${{ github.event.deployment_status.environment_url }} }}
-          PLAYWRIGHT_TEST_USER_PASS: {{ ${{ secrets.PLAYWRIGHT_TEST_USER_PASS }} }}
+          PLAYWRIGHT_TEST_BASE_URL: {{ "${{ github.event.deployment_status.environment_url }}" }}
+          PLAYWRIGHT_TEST_USER_PASS: {{ "${{ secrets.PLAYWRIGHT_TEST_USER_PASS }}" }}
       - uses: actions/upload-artifact@v4
         if: failure()
         with:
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/.github/workflows/playwright.yml
```diff
@@ -7,18 +7,25 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.deployment_status.state == 'success'
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
oudeismetis, line 11: I minor, but important, upgrade I think people often forget to do. 👍

{{cookiecutter.project_slug}}/.github/workflows/playwright.yml
```diff
@@ -7,18 +7,25 @@ jobs:
...
-          PLAYWRIGHT_TEST_USER_PASS:
-            {{ "${{ secrets.PLAYWRIGHT_TEST_USER_PASS }}" }}
+          PLAYWRIGHT_TEST_BASE_URL: {{ "${{ github.event.deployment_status.environment_url }}" }}
+          PLAYWRIGHT_TEST_USER_PASS: {{ "${{ secrets.PLAYWRIGHT_TEST_USER_PASS }}" }}
+      - uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: playwright-report
+          path: ./client/playwright-report/
+          retention-days: 30
```
oudeismetis, line 31: Not for this PR (unless you want it to be), but one issue I observed the other day on a project was that the deployment failed and so we skipped running Playwright because `github.event.deployment_status.state` did not equal success. But the PR was still green because of the previous playwright run.

oudeismetis (approved):

{{cookiecutter.project_slug}}/.github/workflows/playwright.yml
whusterj: Ah, makes sense. I will move this to GH issue for later.

PR #370, opened 2024-12-30
Example of .well-known file, Pillow version update and Heroku missing dependency fix

JSicardiNC:
## What this does

Adds an example on the server template on how to serve a static file from root. Additionally fixes the issue with Pillow outdated version and a missing Heroku dependency on heroku deploy action.

## Checklist
- [ ] Todo 1
- [ ] Todo 2
- [ ] Todo 3

## How to test


{{cookiecutter.project_slug}}/server/{{cookiecutter.project_slug}}/common/views.py
```diff
@@ -1,10 +1,25 @@
...
 
 
 def index(request):
     try:
         return render(request, "index.html")
     except TemplateDoesNotExist:
         return render(request, "core/index-placeholder.html", status=status.HTTP_404_NOT_FOUND)
+
+
+class WellKnownExampleFile(View):
```
paribaker, line 16: FYI I dont think you need to create a whole view 
https://github.com/thinknimble/tn-spa-bootstrapper/blob/c41cfa9ab799833beb5650eea13c5a6dde42d2ce/%7B%7Bcookiecutter.project_slug%7D%7D/server/%7B%7Bcookiecutter.project_slug%7D%7D/common/favicon_urls.py#L15

paribaker (approved):
I'm just wondering if there is a problem when this is not part of the manifest.json or not. I think we can fix it later if it is!

PR #376, opened 2025-01-09
Add a log_errors decorator for background and async functions

whusterj:
## What this does

I've been meaning to add this for a long while. This is a common pattern I use on many projects to wrap things like background tasks that operate outside of Django's default exception handler.

This catches and explicitly logs any exception in the wrapped function, both to the default logger and to Rollbar if it is enabled.

## Usage:

```python
@log_errors
def my_background_task(...):
    # Do something in the background outside of Django's and
    # DRF's request-response context and default exception handler.
    # Any exceptions that happen will be caught and logged as usual
```


oudeismetis (approved):
love it! ❤️

OkayItsMikael:
👏🏾

PR #394, opened 2025-03-22
remove docker container names and switch to system python

paribaker:
## What this does

@Jsicardi  noticed that there was an issue with .venv being broken in the container, the issue is that the .venv file from the host is being copied over to the container. This is different than befoew when pipenv created a python env in a different directory than the project directory so this was not an issue, additionally we were  using the pipfile.lock without a container so this change does that. 

The client had a set name for its container, this was causing a clash because the container was the same for every project. 
Having predictable names helps when running commands!


{{cookiecutter.project_slug}}/docker-compose.yaml
```diff
@@ -19,11 +20,13 @@ services:
       - local_postgres_data_backups:/backups
 
   redis:
+    container_name: {{cookiecutter.project_slug}}_redis
```
whusterj, line 23: I don't know that this is necessary, and may actually be less convenient than what we currently have.

For instance, I am used to running `docker compose up -d postgres redis` in my project directories, and docker compose automatically prefixes those containers with the directory name, which is also the project name. So they don't conflict with other projects (the port numbers conflict though).

paribaker: yea I was adding this mostly because I saw it added to the client and it was causing me an issue because I already had the client one running so I can just remove them all in that case!

whusterj (approved):
Looks good

Issue #382, opened 2025-03-01
Remove Makefile - maybe replace with a justfile


whusterj:

whusterj:
This has been resolved - Makefile has been removed and replaced with a justfile in the template.

PR #395, opened 2025-03-24
Replace Makefile with justfile

whusterj:
## What this does

This PR replaces the project-level `Makefile` with a `justfile` and makes some improvements and updates to commands.

Per #382, we don't really use the commands in the `Makefile`, and they are falling out of date. Since we use `just` for tn-cli, it makes sense to also include a `justfile` with project-level commands in our projects.


{{cookiecutter.project_slug}}/justfile
```diff
@@ -0,0 +1,119 @@
...
+[group('django')]
+django-migrate:
+    cd server && python manage.py migrate
+
+[group('django')]
+django-shell:
+    cd server && python manage.py shell_plus
+
+[group('django')]
+django-runserver:
```
paribaker, line 99: I noticed that the `make logs` is missing from this file

whusterj: I can add that real quick!

PR #398, opened 2025-05-01
Frontend fix: Bad API path to fetch system prompt

paribaker:
## What this does

Fix the API call path for system prompts on the frontend!


whusterj (approved):
Good fix. Thanks!

PR #417, opened 2025-08-12
Agent Types on Prompt Template

JSicardiNC:
## What this does

This PR adds agent type support to the PromptTemplate model, allowing prompt templates to be filtered and associated with specific agent types. This enables different AI agents to use different sets of prompt templates based on their specific roles.

## Changes Made

- Added `AgentType` enum to PromptTemplate model with initial `CHAT` type
- Added `agent_types` ArrayField to store multiple agent types per template
- Updated `get_assembled_prompt()` and `aget_assembled_prompt()` methods to accept an optional `agent` parameter for filtering
- Modified ChatConsumer to specify `CHAT` agent type when fetching prompts
- Updated fixtures to include agent_types field for existing templates
- Added database migration for the new field

## Checklist
- [x] Added AgentType enum with CHAT option
- [x] Implemented agent_types ArrayField on PromptTemplate model
- [x] Updated manager methods to support agent filtering
- [x] Modified ChatConsumer to use agent-specific prompts
- [x] Updated fixtures with agent_types data
- [x] Created migration for schema changes

## How to test

1. Run migrations: `python manage.py migrate`
2. Load updated fixtures: `python manage.py loaddata chat/fixtures/prompt_templates.json`
3. Connect to the chat WebSocket endpoint
4. Verify that only templates with `CHAT` agent type are included in the system prompt
5. Create a new template without agent_types and verify it's not included for CHAT agents
6. Create a new template with `CHAT` in agent_types and verify it's included

### DIFF
--- a/{{cookiecutter.project_slug}}/pytest.ini
+++ b/{{cookiecutter.project_slug}}/pytest.ini
@@ -8,3 +8,4 @@ filterwarnings =
 
 markers =
     use_requests: Allow a unit test to call the internet
+    asyncio: Mark test as using asyncio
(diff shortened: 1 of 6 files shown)


whusterj (commented):
@paribaker These checks are still failing. What can/should we do about this? I will go ahead and merge this for now, but I don't like the red X's 😭 

image

PR #433, opened 2025-09-11
update vite plugin to use react-swc includes eslint upgrades

paribaker:
## What this does

Update the plugin used by vite for compilation to use plugin-react-swc 

Performance: It's significantly faster. Your development server will start quicker, and updates via Hot Module Replacement (HMR) will feel instantaneous. 🚀

Modern Default: It is the current standard for all new Vite + React projects.

Simplicity: It handles all the essential React transformations you need with minimal configuration.


whusterj (approved):
Looks like it's building and e2e tests are passing, so that's good enough for me. Merging now.

PR #436, opened 2025-09-13
adds groups to PR and main merge so that only one build for mobile runs at a time

paribaker:
## What this does

The idea here is to stop triggering builds too frequently as they cost on expo. 

This does not cancel the actual build on expo but it does stop the GH actions that previously started so it will help reduce builds when there are rapid consecutive pushes.


whusterj (approved):
👍🏻

PR #453, opened 2026-02-04
Justfile Improvements

JSicardiNC:
## What this does

Adds modifications to justfile to improve pytest commands

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,9 +23,6 @@ dependencies = [
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 
-[tool.coverage.report]
-fail_under = 20
-
 [tool.ruff]
 line-length = 100
 exclude = [
--- a/{{cookiecutter.project_slug}}/justfile
+++ b/{{cookiecutter.project_slug}}/justfile
@@ -47,7 +47,7 @@ server-run:
 
 [group('server')]
 server-test: django-check-migrations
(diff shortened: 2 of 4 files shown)


{{cookiecutter.project_slug}}/justfile
```diff
@@ -45,8 +46,8 @@ server-run:
     @echo "🚀 Backend ready at http://localhost:8000"
 
 [group('server')]
-server-test:
-    uv run pytest server
+server-test: django-check-migrations
+    cd server && uv run python -m pytest --mccabe --cov={{ "{{ project }}" }} -vv {{ "{{ project }}" }} && coverage report --fail-under=20
```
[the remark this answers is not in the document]
whusterj: Whoops, sorry. I misread it!

`--fail-under=20` is also defined in `{{cookiecutter.project_slug}}/.github/workflows/pytest.yml:50` and there it is a different number (60, not 20).

To unify this, I think we can add the following to `pyproject.toml`:

```toml
  [tool.coverage.report]
  fail_under = 20
```

pyproject.toml
```diff
@@ -23,6 +23,9 @@ dependencies = [
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 
+[tool.coverage.report]
+fail_under = 20
```
whusterj: I think we can bump this up to 60. I think the bootstrapper has 65%+ coverage out of the box.

JSicardiNC: my bad, you are right. I fixed this comment and the 2 above to match with the current dir structure

PR #503, opened 2026-06-12
Fix: Dependency Update

### LINKED ISSUE #497: Upgrade ruff version in linting pipeline
Ruff version in linting.yml pipeline has become pretty outdated, and generates failures in PRs vs the current versions being used in the bootstrapper. Update to version 0.9.8 (the new minimum in pyproject.toml)

### LINKED ISSUE #499: Upgrade Boto minimum version for Pydantic-AI
Several projects use Pydantic-AI with minimum version 0.1.1. Current version is 1.84.1. To start using similar versions, it requires an upgrade of boto dependency to version 1.42.63

JSicardiNC:
## What this does

Routine dependency maintenance for the backend template:

- Bumps **boto3** `1.39.17` → `==1.42.*` (repo-standard pin; pulls transitive **botocore** / **s3transfer**) and regenerates `uv.lock`.
- Runs **ruff** in the linting workflow from the locked env (`uv run ruff` via `astral-sh/setup-uv`) instead of a hard-coded version, so CI lint/format checks can never drift from the ruff we develop against.

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/linting.yml
@@ -5,18 +5,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install ruff
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
       - name: Run Ruff Lint Check
-        run: ruff check --output-format=github server
+        run: uv run ruff check --output-format=github server
       - name: Run Ruff Format Check
-        run: ruff format --check server -q
+        run: uv run ruff format --check server -q
 {%- if cookiecutter.client_app == 'React' %}
   Lint_Web:
     runs-on: ubuntu-latest
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -5,7 +5,7 @@ description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "boto3==1.39.17",
+    "boto3==1.42.*",
     "channels==4.2.*",
     "channels-redis==4.2",
     "daphne>=4.1.2",
--- a/{{cookiecutter.project_slug}}/uv.lock
+++ b/{{cookiecutter.project_slug}}/uv.lock
@@ -230,30 +230,30 @@ wheels = [
 
 [[package]]
 name = "boto3"
-version = "1.39.17"
+version = "1.42.97"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore" },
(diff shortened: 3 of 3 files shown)


{{cookiecutter.project_slug}}/.github/workflows/linting.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install ruff
+          pip install ruff==0.8.6
```
whusterj: Pinning ruff in CI is the right call, but `0.8.6` is actually *below* our declared `ruff>=0.9.8` in `pyproject.toml` (and the lock resolves to `0.14.10`), so CI would format-check with an older ruff than we develop against, and the two can disagree. Rather than hard-code a version here, let's run ruff straight from the locked env so it can never drift from the lockfile:

```yaml
      - name: Install uv
        uses: astral-sh/setup-uv@v5
      - name: Run Ruff Lint Check
        run: uv run ruff check --output-format=github server
      - name: Run Ruff Format Check
        run: uv run ruff format --check server -q
```

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "boto3==1.39.17",
+    "boto3>=1.42.63",
```
whusterj: Let's use `boto3==1.42.*` instead of `>=1.42.63`. That's our standard for most deps (`channels==4.2.*`, `dj-rest-auth==7.0.*`, etc.). This gives us patch versions without inadvertent minor or major updates, which can often bring breaking changes. And honestly today with as many supply chain attacks as there are, it may be best to pin all dep to specific versions, but that's not necessary here.

Make sure to regen uv.lock after making this change.

whusterj (changes requested):
Thanks for the cleanup! See my two comments. I also updated the PR description.

{{cookiecutter.project_slug}}/.github/workflows/linting.yml
JSicardiNC: Oh good call, i actually moved that line from another project, so maybe i will be fixing this on both projects. Thanks

{{cookiecutter.project_slug}}/pyproject.toml
JSicardiNC: good call, will do

paribaker (dismissed):

whusterj:
Rebased onto `main` and implemented the review feedback:
- **linting.yml**: dropped the hard-coded `ruff==0.8.6` (which was below our declared `ruff>=0.9.8`); now runs `uv run ruff check/format` from the locked env via `astral-sh/setup-uv@v5`, so it can't drift from the lockfile.
- **pyproject.toml**: `boto3==1.42.*` (repo standard) and regenerated `uv.lock` (minimal update — boto3 → 1.42.97 + botocore/s3transfer, kept at revision 1).

Verified by rendering the template and running Heroku's exact `uv sync --locked` (uv 0.11.28) — passes. Note the `sign-up.spec.ts` change is now identical to main, so it dropped out of the diff.

PR #551, opened 2026-08-04
Scope the fast-uri override to the major that consumers request

nimble-claude:
## What This Does

Changes `overrides.fast-uri` from `>=4.1.1` to `>=3.1.5 <4.0.0` in the web and the mobile client, and regenerates both lock files. This clears alerts #829 and #830, GHSA-7p8r-x3mc-p8w7 (CVE-2026-18446, high).

The advisory has three vulnerable ranges: `<2.4.4`, `>=3.0.0 <3.1.5`, and `>=4.0.0 <4.1.2`. Both clients resolved fast-uri 4.1.1, which is in the third range.

No package in either client asks for fast-uri 4. Every consumer is an ajv 8.18.0 that declares `^3.0.1`:

| Client | Consumers of fast-uri |
|---|---|
| web | `ajv-formats`, `webpack`, `terser-webpack-plugin` |
| mobile | the same three, plus `expo-build-properties` and `expo-dev-launcher` |

The unbounded `>=4.1.1` override is what put 4.x in the tree. A higher floor on that same major would keep a major that no consumer requested, and give the next 4.x advisory the same effect. The override is now bounded to the 3.x major that the consumers declare. Version 3.1.5 is patched against all five current fast-uri advisories: GHSA-7p8r-x3mc-p8w7, GHSA-v2hh-gcrm-f6hx, GHSA-4c8g-83qw-93j6, GHSA-v39h-62p7-jpjc and GHSA-q3j6-qgpj-74h6.

This repeats the correction made for js-yaml in 706a0ea, where an override onto an unrequested major was itself the cause of the exposure.

Only fast-uri changes. The lock diff moves one hoisted `node_modules/fast-uri` 4.1.1 entry to nested 3.1.5 entries, one for each consumer. No other package moves.

## Note for reviewers

Check that no future dependency needs fast-uri 4. The `<4.0.0` bound is deliberate: it holds the override inside the major that ajv 8 declares. If a package later requests fast-uri 4, this bound blocks it and the install fails, which is a visible error and not a silent downgrade. Raise the bound at that time.

Verification, all in a `node:22` Linux container on a project generated from this branch with `cookiecutter/rn_template.yaml`:

- `npm ci` in `client` and in `mobile`: both succeed. This proves the committed lock files agree with the changed `package.json`.
- Web: `npm run tslint` (tsc) exit 0, `npm run eslint --max-warnings=0` exit 0, `npm run build` exit 0, `vitest run` 1 test passed.
- Mobile: `npm run lint:tslint` exit 0, `npm run lint:eslint --max-warnings=0` exit 0, and `npx expo export --platform web` exit 0.
- All 8 installed copies of fast-uri on disk report 3.1.5, in both clients.

The e2e Playwright specs were excluded from the vitest run. Vitest collects them and they fail without a server; that predates this change.

> Do not merge this PR without human review.


whusterj (approved):
Approved. This is the root-cause fix: it removes the unbounded fast-uri override that pulled the unrequested 4.x major into the tree, and scopes the override to the 3.x major that every consumer (ajv 8) declares. 3.1.5 is patched against all current fast-uri advisories. Required CI checks pass; the only red check (publish) fails for an unrelated expired iOS provisioning profile. Supersedes #552.

